PR #1, opened 2019-10-24
Add foundation of express and create-react-app

synzen:
- Added [gitignore](https://github.com/nodejs/node/blob/master/.gitignore)
- Updated readme for instructions
- Added express
- Added create-react-app (with proxy to express)

feel free to make changes to match whatever workflow you're comfortable with (this PR might've been too big)


synzen:
Too big - splitting it up

PR #3, opened 2019-10-24
Set up express.js

synzen:


synzen:
I can move it to a folder but what would the folder be for? most express apps have the entry file (server.js in this case ) in the root dir

MicLi01:
> I can move it to a folder but what would the folder be for? most express apps have the entry file (server.js in this case ) in the root dir

By moving the express code to its own directory, it will be clear what each module is doing. In the root directory we can add a package.json file which would have scripts to run the client and the express server.

PR #5, opened 2019-10-25
Update README.md

oscarchong:
updated basic idea of our app on readme


yliu0715:
@oscarchong I think you forgot a space between "##" and Purpose. Can you double check that?

oscarchong:
> @oscarchong I think you forgot a space between "##" and Purpose. Can you double check that?

fixed and added more info

PR #6, opened 2019-10-25
database set up

MicLi01:


synzen:
I'm guessing the route you put in `server/routes/basicroute` is supposed to return the results from the test model right?

MicLi01:
> 
> 
> I'm guessing the route you put in `server/routes/basicroute` is supposed to return the results from the test model right?
That's right. You can add it into the sever index.js file to test that your database worked

server/routes/basicroute.js
```diff
@@ -0,0 +1,28 @@
...
+      console.log('Failure to retrieve');
+      return 'Fail';
+    }
+    else {
+      console.log(test_res);
+      return test_res;
+    }
+    res.redirect('/');
+  });
+});
```
synzen: i think this is what you meant to do in that case

```js
router.get('/testquery', (req, res) => {
  test.find({}, function(err, test_res) {
    if (err) {
      return next(err);
    }
    console.log(test_res);
    res.json(test_res);
  });
});
```

or better with promises
```js
router.get('/testquery', (req, res) => {
  test.find({}).then(test_res => {
    console.log(test_res);
    res.json(test_res);
  }).catch(next)
});
```

cause returning the results within the test.find callback won't be visible to the router to send back the response without using `res.json` for json, `res.text` for plaintext, or w/e you want

the `next` function will auto-pass the error to the express error-handler and log it

MicLi01: I didn't know that. Nice catch. I made the changes.

MicLi01:
> 
> 
> I think you can change this to use promises just to stay consistent with the other changes.
> 
> ```
> mongoose.connect('mongodb://localhost:27017/quikborrow', { useUnifiedTopology:true, useNewUrlParser: true})
> .then(() => console.log('Successfully connected to database'))
> .catch(err => console.log('Database Connection error', err));
> ```
addressed

PR #8, opened 2019-10-26
Add server gitignore

synzen:
Forgot gitignore for server - otherwise git doesn't ignore node_modules in there


oscarchong:
Good catch

PR #12, opened 2019-10-26
User reg auth

Akasora39:


synzen:
looks like you may have missed the path changes in the merge (`client/src/js/components` rather than `client/src/components`) for where components go (also the imports in App.js)

also the server dependencies were installed in the root package.json rather than [server/package.json](https://github.com/CSC59939/QuickBorrow/blob/master/server/package.json) - not sure if this is what you intended

package.json
```diff
@@ -6,12 +6,25 @@
...
+    "bcryptjs": "^2.4.3",
+    "body-parser": "^1.19.0",
+    "express": "^4.17.1",
+    "is-empty": "^1.2.0",
+    "jsonwebtoken": "^8.5.1",
+    "mongoose": "^5.7.7",
+    "passport": "^0.4.0",
+    "passport-jwt": "^4.0.0",
+    "validator": "^11.1.0"
+  },
```
yliu0715: server dependencies should be in the server directory and client dependencies should be in client directory

client/src/js/components/App.js
```diff
@@ -2,6 +2,16 @@ import React from 'react';
...
+<<<<<<< HEAD
+import Homepage from './Homepage/Homepage';
+import Register from './Register/Register';
+import Login from './Login/Login';
+=======
+import Homepage from './components/Homepage/Homepage';
+import Register from './components/Register/Register';
+import Login from './components/Login/Login';
+>>>>>>> 443c33d4f839c5178919d3de0b6ed13bc591e5de
+
```
yliu0715: Did you create a folder for each individual component? I think you might want the lower part here

Akasora39: Sorry about that, merging issue that I forgot to catch after pushing.

PR #14, opened 2019-10-27
added DetailedPage component

oscarchong:


yliu0715:
What is this component for?

oscarchong:
This is a landing page for user after they search and click on the item, it's to show the full detail of the item to which allows them to make a request "borrow"

yliu0715:
@oscarchong Ahh I see. Hold off on this for now, because I'm currently working on the products list page and this component that you setup is relevant to what I'm doing. Can you work on the search functionality in the meantime? You can create a component for the search bar just like this one too.

synzen:
I was also working on the lending page but it sounds like what you're doing is similar to this so I'll halt progress on that

PR #18, opened 2019-10-28
Add relevant CSS files, replace App.js contents

synzen:
- Home page has very basic routing to demonstrate the different components
- Missing CSS files from primereact are added

### DIFF
--- a/client/src/js/components/App.js
+++ b/client/src/js/components/App.js
@@ -1,28 +1,44 @@
-import React from 'react';
-import logo from '../../images/logo.svg';
+import React, { useState } from 'react';
 import '../../css/App.css';
-
+import 'primereact/resources/themes/nova-light/theme.css';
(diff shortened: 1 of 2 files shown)


client/src/js/components/App.js
```diff
@@ -1,28 +1,44 @@
...
+    case 1:
+      toRender = <Register />
+      break
+    case 2:
+      toRender = <Login />
+      break
+    default:
+      toRender = <div>No Page</div>
+  }
+
```
yliu0715: @synzen  Are you trying to render different routes here? We can use react-router-dom for this. It's a great library for specifying the client routes.

synzen: yeah it was very basic just to demonstrate the different routes for everyone working on the different components, without having to set up react-router at the moment

PR #21, opened 2019-10-28
Fix npm install breaking with call stack size exceeded

synzen:


JackyLiu97:
beast. i'll merge it to master then.

PR #25, opened 2019-10-30
User reg backend

Akasora39:
- Set up user db model 
- Set up user registration and login api endpoints
- Set up testing environment for client and server
- Added tests for user model and api endpoints(still has issues)
- Modified linting to be run on both client and server simultaneously


yliu0715:
Bro, I know my PR was big but this is next level.

synzen:
too big pr

Akasora39:
The main files of interest are the ones involved with the user registration and authentication. You can ignore the config files and tests as they were moreso meant to give you guys an idea of how our tests should look like.

server/index.js
```diff
@@ -1,8 +1,36 @@
...
+const db = require("./config/keys").mongoURI
+// Connect to MongoDB
+mongoose
+  .connect(
+    db,
+    { useNewUrlParser: true }
+  )
+  .then(() => console.log("MongoDB successfully connected"))
+  .catch(err => console.log(err))
+
```
yliu0715: @Akasora39 We already have the database setup

server/index.js
```diff
@@ -1,8 +1,27 @@
...
+const port = process.env.PORT || 8081 // process.env.port is Heroku's port
+const users = require("./routes/api/users")
+
+// Bodyparser middleware
+app.use(
+  bodyParser.urlencoded({
+    extended: false
+  })
+)
+app.use(bodyParser.json())
```
synzen: the bodyparser module is no longer needed since express has these middleware functions built in now

```js
app.use(express.json())
app.use(express.urlencoded())
```

Akasora39: I wasn't aware of this but it really does simplify the whole process. Thanks for letting me know. We can start with unit tests. I was more curious about the integration testing first and was caught up on it because of discrepancies between my results with Postman.

synzen: np - yeah, but not with supertest b/c supertest spins up the actual server where the app resolves the URL you gave it. 

to unit test route functions, they should ideally be exported from the file

```js
// file.js
function getFunction(req, res) {}

app.get('/api/hello', getFunction)

module.exports = {
  routes: {
    getFunction
  }
}
```
so that the route handler can be specifically targeted for testing
```js
describe('Unit: /api', () => {
  describe('GET /hello', () => {
    it('works', () => {
      file.routes.getFunction(mockRequest, mockResponse)
      // test mockRequest, mockResponse
    })
  })
})
```

something like that - but the integration tests you wrote are totally fine - was just curious if we're going straight of int testing lol

server/routes/api/users.js
```diff
@@ -0,0 +1,112 @@
...
+			  bcrypt.genSalt(10, (err, salt) => {
+				bcrypt.hash(newUser.password, salt, (err, hash) => {
+				  if (err) throw err;
+				  newUser.password = hash;
+				  newUser
+					.save()
+					.then(user => res.json(user))
+					.catch(err => console.log(err));
+				});
+			  });
```
MicLi01: is new line on newUser.save intended?

server/routes/api/users.js
```diff
@@ -0,0 +1,112 @@
...
+		.isLength({ min: 6 })
+		.custom((value,{req, loc, path}) => {
+            if (value !== req.body.password2) {
+                throw new Error("Passwords don't match");
+            } else {
+                return value;
+            }
+        })
+	], (req, res) => {
+		const errors = validationResult(req);
```
MicLi01: Seems to be a lot going on in this file. You could abstract away the functionality into the handlers directory I created. But there's no need to do that right now.

server/index.js
```diff
@@ -4,7 +4,19 @@ const router = require("./routes");
...
+// Bodyparser middleware
+app.use(express.urlencoded({ extended: false }))
+app.use(express.json())
+  
+// Passport middleware
+app.use(passport.initialize())
+// Passport config
+require("./config/passport")(passport)
+// Routes
+app.use("/api/users", users)
```
yliu0715: I think you took out `app.use("/server", router);` here. Also, let's stay consistent with either `/api/*` or `/server/*` for the route handlers.

server/index.js
```diff
@@ -4,7 +4,19 @@ const router = require("./routes");
...
```
yliu0715: Please stay consistent with either `/api/*` or `/server/*` for the route handlers.

PR #28, opened 2019-11-02
Fix test, add @types/jest

synzen:
@types/jest allows intellisense in vscode to detect jest

the Hello World test failed due to misconfiguration - this fixes it


synzen:
fixed~~

PR #29, opened 2019-11-03
Set up continuous integration

synzen:
- github actions nodejs workflow file added
- added necessary lint scripts to package.jsons
- ` --detectOpenHandles --forceExit` were added to the jest script because of an issue of the index.test never actually closing - see https://github.com/visionmedia/supertest/issues/520. so this is just a workaround
- and an unrelated fix to make the index.test cleaner


yliu0715:
Is the Server ESLint check failing because of the fact that Unix and Windows systems have different linebreaks?

yliu0715:
This is neat!

synzen:
ikr - fuhgedabbout travisCI/circleCI, github actions is where its at - easy to setup and get started with 🙌 🙌 🙌 🙌 

yeah, looks like you're right that that's one of the reasons - https://stackoverflow.com/questions/37826449/expected-linebreaks-to-be-lf-but-found-crlf-linebreak-style

but we have 611 problems in total with our server eslint lol

https://github.com/CSC59939/QuickBorrow/pull/29/checks?check_run_id=286336189

image

On another note, the Marketplace for github actions has a bunch of cool stuff but haven't touched them yet https://github.com/marketplace?type=actions

Akasora39:
Eslint was configured using recommended rules so there may be some things being checked that we don't necessarily want. 
The rules with check marks are applied with recommended: https://eslint.org/docs/rules/
We can change the configuration if needed. I think the only thing we explicitly wanted was single quote consistency.

synzen:
I'm fine with any rules so long as we have standards tbh, but the scope of this PR is to just set up the CI process

after this we can just use `--fix` to autofix most of them, and manually address the rest of the issues

PR #37, opened 2019-11-07
Add item route unit tests, add server scripts

synzen:
- Add item route unit tests
- Organize tests into test dir
- Add separate scripts for unit and integration runs (since int runs are slower)
- Add `lint-fix` script

### DIFF
--- a/server/database/models/User.int.test.js
+++ b/server/database/models/User.int.test.js
@@ -1,7 +1,7 @@
 const mongoose = require('mongoose');
-const UserModel = require('../../../database/models/User');
+const UserModel = require('./User');
 const userData = {name: 'Test', email: 'test@test.com', password: 'password'};
-const keys = require('../../../config/keys');
+const keys = require('../../config/keys');
 const MONGODB_OPTIONS = {
   useNewUrlParser: true,
   useCreateIndex: true,
--- a/server/index.int.test.js
+++ b/server/index.int.test.js
@@ -1,8 +1,8 @@
-const app = require('../index');
+const app = require('./index');
 const supertest = require('supertest');
 const request = supertest(app);
 const mongoose = require('mongoose');
-const keys = require('../config/keys');
+const keys = require('./config/keys');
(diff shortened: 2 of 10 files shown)


synzen:
we can do that - changed accordingly

I was also looking through the tests and found this one https://github.com/CSC59939/QuickBorrow/blob/2e558a68f2d9152b2d7b82aef0452b494da88945/server/database/models/User.int.test.js

i don't believe this would be a relevant integration test because this is testing _mongoose_ and _mongodb_ functions, rather than the function of our own logic (and such functions would already be tested by the mongodb/mongoose teams themselves)

Akasora39:
> we can do that - changed accordingly
> 
> I was also looking through the tests and found this one https://github.com/CSC59939/QuickBorrow/blob/2e558a68f2d9152b2d7b82aef0452b494da88945/server/database/models/User.int.test.js
> 
> i don't believe this would be a relevant integration test because this is testing _mongoose_ and _mongodb_ functions, rather than the function of our own logic (and such functions would already be tested by the mongodb/mongoose teams themselves)

That's true. I suppose the file in question we really want to be testing would be users.js in the routes directory. Also, going off from this, that would mean we wouldn't be testing express-validator for our integration tests.

synzen:
yea, since we're using express-validator in integration tests, i believe we just have to assert that if we send an invalid json body in the request, then the response should have an indication showing what was wrong in the body (basically test the public interface, instead of the internals that is express-validator). we're working under the assumption that express-validator works as intended, and returns the appropriate response (CSC 335 paradigms throwback!)

Akasora39:
> yea, since we're using express-validator in integration tests, i believe we just have to assert that if we send an invalid json body in the request, then the response should have an indication showing what was wrong in the body (basically test the public interface, instead of the internals that is express-validator). we're working under the assumption that express-validator works as intended, and returns the appropriate response (CSC 335 paradigms throwback! if you know what I mean lol)

Lol alright sounds good.

PR #40, opened 2019-11-07
Revert "Add item route unit tests, add server scripts"

MicLi01:
I don't agree with all the things in this pull request. I believe the routes should be in a separates directory split from the functionality.


synzen:
looks like you're probably referring to PR https://github.com/CSC59939/QuickBorrow/pull/34 - reverting this one will just remove the unit tests

MicLi01:
> 
> 
> looks like you're probably referring to PR #34 - reverting this one will just remove the unit tests (I'm fine with whatever btw)

The unit tests looked fine to me. I had an issue with adding in the routes into the same file as the functionality. The routes should be in a different file to make it as abstract as possible. I also have an issue with how everything was moved into the routes directory. The functionality of getting items and creating items don't really have much to do with the routes themselves but this wasn't part of this pull request.

MicLi01:
> 
> 
> > looks like you're probably referring to PR #34 - reverting this one will just remove the unit tests (I'm fine with whatever btw)
> 
> The unit tests looked fine to me. I had an issue with adding in the routes into the same file as the functionality. The routes should be in a different file to make it as abstract as possible. I also have an issue with how everything was moved into the routes directory. The functionality of getting items and creating items don't really have much to do with the routes themselves but this wasn't part of this pull request.

The second thing I mentioned can probably be fixed with a bit of renaming of the directories though.

PR #42, opened 2019-11-07
front end user profile

JackyLiu97:
Creating a user profile that includes, wishlist for user to wish to borrow, item list that show items the user have to sell/lend, and personal information like pic, age, school, major, name, etc.


JackyLiu97:
Okay I will add a photo of themselves

client/src/js/components/App.js
```diff
@@ -21,6 +23,7 @@ class App extends Component {
             <Route exact path="/register" component={Register} />
             <Route exact path="/login" component={Login} />
             <Route exact path="/products" component={ProductList} />
+            <Route exact path="/profile" component={Profile} />
```
yliu0715: What you are doing here is you created a static route for profile. But you can imagine that each user will render a different profile page based on their unique user content. So what you might want to look into is creating a dynamic route based on a unique user identifier.
Here is a link that might prove to be useful: https://learnwithparam.com/blog/dynamic-pages-in-react-router/

JackyLiu97: Thank you I will look into that

JackyLiu97:
Okay I will write a unit test for it as well

client/src/js/components/App.js
JackyLiu97: I'll set up the API call for User Profile in another PR. this is just to mainly set up the page.

Issue #43, opened 2019-11-08
Server Structure


synzen:
if we're to separate the route handler from the routes, shall we follow the model/routes/controller/services structure then?

reference:
https://riptutorial.com/node-js/example/32331/model-routes-controllers-services-directory-structure
https://riptutorial.com/node-js/example/32332/model-routes-controllers-services-code-structure

MicLi01:
That structure looks good to me

PR #44, opened 2019-11-08
Refactor server for models/controller/services/routes/middleware

synzen:
addresses #43 but the users route is left unchanged for an easier review

see https://github.com/CSC59939/QuickBorrow/tree/mt/arch-refactor/server for the structure

- Moved `/database/index` to `/services/mongodbConnect`
- Moved `/database/models` to `/models`
- Added 
  - `/controllers` - Route handler functions
  - `/services` - The business/CRUD logic of controllers
  - `/middleware` - Middleware functions (functions that are in the "middle" of when express receives the request, and when the controllers in `/controllers` handle them)

let me know what you guys think


JackyLiu97:
This a nice quality of life change to the server directory. approved !

yliu0715:
Another thing we can consider is to not commit the key file at all. We can import one set of key files for production call it `prodKeys.js` and another set of key for development call it `devKeys.js`. This way, everyone has their own set of keys for local development, and the production keys for, yes, production.

synzen:
yeah I found myself changing the mongo URI to have an easier time to test the API routes so a solution to that would be good - though that can be a different PR/issue since this one's focused on changing the folder structure

PR #46, opened 2019-11-08
updated homepage

oscarchong:
- Created Logo for our product

- Added some styling


client/src/js/components/Homepage/Homepage.js
```diff
@@ -1,17 +1,35 @@
...
+`
 
 class Homepage extends Component {
 	render() {
 		return (
 			<div>
-				<div>
-					<h4>quiKBorrow</h4>
+				<Center>
+				<img src={quikLogo}/>
```
yliu0715: This is minor, but you can add in an alt tag in case logo doesn't render.

PR #48, opened 2019-11-09
unit tests for product stuff

yliu0715:


yliu0715:
> Looks good. Just wanted to let you know about a workaround I figured out for the mock store. Instead of testing the component connected to the store, you can just export the unconnected component. This makes testing faster and simpler. In your case, this would be an integration test since it goes from component to action to reducer to store.

Thanks, I'll keep that in mind for the next tests.

client/src/js/actions/index.js
```diff
@@ -5,15 +5,15 @@ import { FETCH_PRODUCTS, FETCH_PRODUCT } from "./types";
...
 };
 
 // Get one product
 export const fetchProduct = itemId => {
   return async function(dispatch) {
-    const res = await axios.get("/server/items/retrieve", {
+    const res = await axios.get("/api/item", {
       itemId: itemId
     });
     dispatch({ type: FETCH_PRODUCT, payload: res.data });
```
synzen: the URL would actually be
```js
const res = await axios.get(`/api/items/${itemId}`);
```
https://github.com/CSC59939/QuickBorrow/blob/fb6c1d85bb5bc27de88fea01710acd62d486eb44/server/routes/api/items.js#L25
following [REST](https://restfulapi.net/resource-naming/) principles

client/src/js/reducers/productsReducer.test.js
```diff
@@ -0,0 +1,30 @@
...
+                }]
+            })).toEqual([
+                {
+                    availability: true,
+                    description: "test description",
+                    name: "test name",
+                    user: "5dbcedbc2f01110000da8ac8",
+                    _id: "5dbcedae1c9d4400009df4a5" 
+                }
+            ])
```
synzen: a minor thing but this can be reduced if you want
```js
const payload = {
	availability: true,
	description: "test description",
	name: "test name",
	user: "5dbcedbc2f01110000da8ac8",
	_id: "5dbcedae1c9d4400009df4a5" 
}
productsReducer([], {
	type: types.FETCH_PRODUCTS,
	payload: [{ ...payload }]
})).toEqual([payload])
```

client/src/js/actions/index.js
yliu0715: That was actually my initial URL, but I received a 422 error. I then checked the server and it seems like `itemId` is being extracted from the request body. Hence why I changed it to that. Can you check on this for me?

client/src/js/reducers/productsReducer.test.js
yliu0715: Yes, that is definitely cleaner!

client/src/js/actions/index.js
synzen: yessir you were right, 
image

this is actually a bug I introduced, sorry bout that:

https://github.com/CSC59939/QuickBorrow/blob/fb6c1d85bb5bc27de88fea01710acd62d486eb44/server/routes/api/items.js#L26

the error says invalid value because I forgot the id is alphanumeric as per mongodb IDs, and was checking exclusively for integers. if I change `.isInt()` to `isAlphanumeric()`, then it works as expected:
image
thanks for the heads up, #52 should fix it

yliu0715: sweet

yliu0715: Changes have been addressed.

PR #49, opened 2019-11-09
User reg redux

Akasora39:
- Finish redux setup of user registration allowing users to connect to the store and successfully register and login
- Setup dashboard which users will be redirected to upon login 
- Add tests for reducers
- Update tests for registration and login

Sorry for the length of this one. I tried to keep it small but connecting to redux involved a lot of changes for various files.


yliu0715:
But you could've created a branch for the feature implementation and a separate branch for the tests.

Akasora39:
> But you could've created a branch for the feature implementation and a separate branch for the tests.

Yeah but the existing tests weren't passing because of my changes so I had to fix that. In retrospect, I probably could've had some default test that would always pass but I felt like I should include the test relevant to the components in the same branch.

synzen:
I'm going to get a lush head of beautiful gray hair by the time I finish this review

but I do think that it could've been broken up by the redux additions, and then after that's merged, the component additions alongside their respective tests - if you already made so many changes on a branch, you could commit & push the relevant ones, checkout to another branch and push some there and etc

client/src/js/components/App.js
```diff
@@ -21,6 +23,9 @@ class App extends Component {
             <Route exact path="/register" component={Register} />
             <Route exact path="/login" component={Login} />
             <Route exact path="/products" component={ProductList} />
+            <Switch>
+              <PrivateRoute exact path="/dashboard" component={Dashboard} />
```
synzen: I think the PrivateRoute logic can just be put in App.js - the way it is now, it kind of increases the complexity and reduces the clarity of the routing

Above the return statement, just a simple

```js
if (this.props.auth.isAuthenticated === true) {
  return <Redirect to='/login' />;
}
// the rest stays the same, beside PrivateRoute ofc
```

Akasora39:
> I'm going to get a lush head of beautiful gray hair by the time I finish this review
> 
> but I do think that it could've been broken up by the redux additions, and then after that's merged, the component additions alongside their respective tests - if you already made so many changes on a branch, you could commit & push the relevant ones, checkout to another branch and push some there and etc

I was trying to have the user registration and authentication feature working so that the user profile part could be implemented by building off of that so I wanted to get that working as soon as possible. I do see your point though and will keep this in mind for the future.

client/src/js/reducers/authReducer.test.js
```diff
@@ -0,0 +1,38 @@
...
+		const user = {
+			name: 'Test',
+			email: 'test@test.com'
+		}
+		const newState = authReducer(initialState, {
+			type: SET_CURRENT_USER,
+			payload: user
+		})
+		expect(newState.user).toEqual(user)
+	})
```
synzen: for a more robust test, a copy of the user object should be passed to the reducer:

```js
const newState = authReducer(initialState, {
	type: SET_CURRENT_USER,
	payload: { ...user }
})
```
because if passed the original object and the reducer function unexpectedly changes the user's name for example, the test will still pass because the original user object was modified, which is also attached to `newState.user` - and the assertion will always pass even though it shouldn't

client/src/js/components/App.js
Akasora39: I'm not sure I agree with that because this would involve passing props to the App.js file and having a mapStatetoProps function and making it a connected component as well which seems unnecessary.

client/src/js/components/Login/Login.js
```diff
@@ -38,7 +59,14 @@ class Login extends Component {
...
 							id="email"
 							type="email"
+							className={classnames("", {
+                    			invalid: errors.email || errors.emailnotfound
+                  			})}
 						/>
+						<span>
+		                  {errors.email}
+		                  {errors.emailnotfound}
+		                </span>
```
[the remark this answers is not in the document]
Akasora39: I think I'll hold off on making this change until https://github.com/CSC59939/QuickBorrow/pull/50 gets merged because I'm getting some weird behavior when trying to JSON stringify the error where it displays fine until I try to access an element in the array at which point it becomes undefined.

client/src/js/components/App.js
synzen: oh nevermind - I took a closer look at and there were other issues. But I was looking for a way to remove `PrivateRoute` completely since it feels like a hacky fix imo

What do you think about just leaving it as
```js
<Route exact path="/dashboard" component={Dashboard} />
```
and putting 
```js
// Dashboard
  componentDidMount() {
    if (this.props.auth.isAuthenticated !== true) {
      this.props.history.push('/login')
    }
  }
```
in Dashboard - this seems to accomplish the exact same thing, while completely removing the need for the PrivateRoute component

Akasora39: We could do that. I figured it would be useful to have a PrivateRoute component so it could be reused for other components like the user profile but thinking back on it we probably don't need a user to be authenticated to view a profile.

synzen: ohh gotcha I see why you wanted it then

client/src/js/reducers/errorReducer.test.js
```diff
@@ -0,0 +1,21 @@
...
+	})
+
+	it('should return errors', () => {
+		const error = {
+			error: 'Test error'
+		}
+		const newState = errorReducer({}, {
+			type: GET_ERRORS,
+			payload: error
+		});
```
synzen: one more thing, just a clone of error instead of the original error object needs to be passed in `{ ...error }`

client/src/js/components/Register/Register.test.js
```diff
@@ -3,25 +3,58 @@ import renderer from 'react-test-renderer';
...
+	const params = {
+		registerUser: registerUser,
+		auth: {
+			isAuthenticated: false,
+			user: {},
+			loading: false
+		},
+		errors: {}
+	}
+	let wrapper = shallow(<Register registerUser = {params.registerUser} auth = {params.auth} errors = {params.errors} />);
```
synzen: this wrapper should ideally be rewritten for every test case, otherwise the same component unit is shared across multiple tests, which means one unit test has the potential to affect every other unit test that uses this component unit

client/src/js/components/Register/Register.test.js
```diff
@@ -3,25 +3,58 @@ import renderer from 'react-test-renderer';
 import { MemoryRouter } from 'react-router-dom';
 import Enzyme, { shallow } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
-import Register from './Register';
+import { Register } from './Register';
+import { registerUser } from "../../actions/index";
```
synzen: if these are meant to be unit tests, the import should be mocked
```js
jest.mock('../../actions/index')
```
otherwise it would be an integration test since the api calls are included in the original, and passing in the original will cause actual API calls

client/src/js/components/Login/Login.test.js
```diff
@@ -1,16 +1,41 @@
...
+	const params = {
+		loginUser: loginUser,
+		auth: {
+			isAuthenticated: false,
+			user: {},
+			loading: false
+		},
+		errors: {}
+	}
+	let wrapper = shallow(<Login loginUser = {params.loginUser} auth = {params.auth} errors = {params.errors} />);
```
synzen: the wrapper should be recreated for every unit test to make sure multiple tests aren't sharing the same variables

client/src/js/components/Login/Login.test.js
```diff
@@ -1,16 +1,41 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
 import { MemoryRouter } from 'react-router-dom';
-import Login from './Login';
+import Enzyme, { shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { Login } from './Login';
+import { loginUser } from "../../actions/index";
```
synzen: if these are meant to be unit tests, the import should be mocked
```js
jest.mock('../../actions/index')
```
actual api calls would be made if any test triggers it

client/src/js/components/Register/Register.test.js
```diff
@@ -3,25 +3,58 @@ import renderer from 'react-test-renderer';
...
+		expect(wrapper.state('password2')).toEqual('password');
+	});
+
+	test('successful registration', () => {
+		wrapper.find('input[type="text"]').simulate('change', {target: {id: 'name', value: 'test'}});
+		wrapper.find('input[type="text"]').simulate('change', {target: {id: 'email', value: 'email@email.com'}});
+		wrapper.find('input[type="text"]').simulate('change', {target: {id: 'password', value: 'password'}});
+		wrapper.find('input[type="text"]').simulate('change', {target: {id: 'password2', value: 'password'}});
+		wrapper.find('form').simulate('submit', {preventDefault() {}});
+	});
```
synzen: best way to get passing tests every time is to not assert anything within the tests lol

client/src/js/components/Login/Login.js
```diff
@@ -38,7 +59,14 @@ class Login extends Component {
 							error={errors.email}
 							id="email"
 							type="email"
+							className={classnames("", {
+                    			invalid: errors.email || errors.emailnotfound
+                  			})}
 						/>
```
yliu0715: @Akasora39 Looks good but something you might want to look into is redux form. Redux form can be used to store all application form state in a reducer. So it can:

- store all fields that are in the form
- verifies if the field inputs are valid
- Do async validation

Reason I'm suggesting this is because I see that you had to create your own Error actions, error reducers, and errorReducer tests. But Redux form abstracts all that headache for you. If what you have right now works, then we can leave it as is, but this can be a small improvement we can make in the future.

client/src/js/components/Register/Register.test.js
Akasora39: I was debating whether or not to even keep this test as I'm not sure if this is where I should draw the line between whether I test it or that it's already been tested by React.

client/src/js/components/Login/Login.js
Akasora39: Interesting. Thanks for letting me know. I think I'll hold off on doing that in this pr and refactor in a future pr.

client/src/js/components/Register/Register.test.js
Akasora39: Do you have any suggestions on what I should assert? From what I've seen, the only way to test this is to either have a state that indicates whether a submission was made or pass a function as a prop to mock but both seem a bit excessive for a single test.

synzen: you could assert that `registerUser` has been called:
```js
expect(wrapper.prop('registerUser')).toHaveBeenCalled()
```
if you do this, it would be best to `jest.resetAllMocks()` in a `afterEach()` function, assuming that you `jest.mock('../../actions/index')` so any calls to `props.registerUser` in other tests doesn't erroneously cause this one to pass

Akasora39: Ok so I had some issues with expect(wrapper.prop('registerUser')).toHaveBeenCalled() giving me undefined but I managed to get it to work using wrapper.instance().props.registerUser instead.

synzen: nice good to know thanks

Akasora39:
> beautiful x2

LOL. Wanted to wait and make sure there were no other issues before resolving the merge conflict.

PR #53, opened 2019-11-10
Yl/product details

yliu0715:
- Basic single product detail page
- Dynamic routing for single product
- Product Reducer to fetch a single product based on itemId

Still need styling which I suck at.


client/src/js/components/Products/ProductCard.js
```diff
@@ -1,30 +1,37 @@
...
 import { Button } from "primereact/button";
 
 const ProductCard = props => {
-  const { name, user, description, availability } = props;
+  const { id, name, user, description, availability } = props;
 
-  const header = <img alt="product_image" src="image/path/goes/here" />;
+  const header = (
+    <Link to={`/products/${id}`}>
+      <img alt="product_image" src="image/path/goes/here" />
```
synzen: for images you could use https://placeholder.com/ to generate a custom placeholder image of the perfect size if you think it'd be useful

client/src/js/components/Products/ProductDetail.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import * as actions from "../../actions";
+
+class ProductDetail extends Component {
```
Akasora39: Remember that as you unit test, you'll want to export the unconnected component so you should have `export class ProductDetail`. For now this is good, but keep in mind that you have to make this change for unit testing.

yliu0715: Oh yes, you are right. Lol

PR #54, opened 2019-11-10
added renting

MicLi01:
- added in image field in Items model
- added basic renting functionality (will add on to this later)

### DIFF
--- a/server/controllers/items.js
+++ b/server/controllers/items.js
@@ -4,7 +4,7 @@ const create = async (req, res, next) => {
   const data = {
     name: req.body.name,
     description: req.body.description,
-    image: req.body.image
+    image: req.body.image,
   };
   itemServices.createItem(data)
       .then((item) => {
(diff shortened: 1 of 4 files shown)


server/controllers/items.js
```diff
@@ -28,8 +29,18 @@ const getAll = (req, res, next) => {
       .catch(next);
 };
 
+const rent = async (req, res, next) => {
```
oscarchong: If you're using VS code, you can add on an "eslint" extension that can help you fix all the eslint errors.

server/routes/api/items.js
```diff
@@ -27,4 +27,13 @@ router.get('/:itemId', [
   validatorErrors,
 ], itemController.get);
 
+/**
+ * @memberof module:api/items
+ * @name POST /rent
+ */
+router.get('/rent/:itemId', [
+  param('itemId').isInt(),
```
synzen: slight fix, should be `isAlphanumeric()` (#52 )

server/controllers/items.js
MicLi01: good to know

server/routes/api/items.js
MicLi01: nice catch. correction has been made.

PR #57, opened 2019-11-11
Add unit tests for items controller and services

synzen:
As a side note, something I just realized the `GET /api/items/:itemId/rent` route could be replaced by a generic `PATCH /api/items/:itemId` to just edit any keys of an Item, which includes availability - otherwise we'll have any make more routes to edit other keys of an Item

### DIFF
--- a/server/controllers/items.unit.test.js
+++ b/server/controllers/items.unit.test.js
@@ -1,0 +1,47 @@
+const itemsController = require('./items');
+const itemServices = require('../services/items');
+
+jest.mock('../services/items');
+
+describe('controller/items', function() {
+  describe('create', function() {
+    it('returns the created item', async function() {
+      const request = {body: {}};
+      const response = {json: jest.fn()};
+      const item = {foo: 'bar'};
+      itemServices.createItem.mockResolvedValueOnce(item);
+      await itemsController.create(request, response);
+      expect(response.json).toHaveBeenCalledWith(item);
+    });
+  });
+  describe('get', function() {
+    it('returns the item', async function() {
+      const request = {params: {itemId: '2qw3eit8ghj'}};
+      const response = {json: jest.fn()};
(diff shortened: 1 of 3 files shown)


MicLi01:
> 
> 
> As a side note, something I just realized the `GET /api/items/:itemId/rent` route could be replaced by a generic `PATCH /api/items/:itemId` to just edit any keys of an Item, which includes availability - otherwise we'll have any make more routes to edit other keys of an Item

That is good to know. Will keep that in mind.

server/controllers/items.unit.test.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+  describe('rent', function() {
+    it('returns the items', async function() {
+      const request = {params: {itemId: '2qw3eit8ghj1'}};
+      const response = {json: jest.fn()};
+      const item = {ab: 'ba'};
+      itemServices.rentItem.mockResolvedValueOnce(item);
+      await itemsController.rent(request, response);
+      expect(response.json).toHaveBeenCalledWith(item);
+    });
```
MicLi01: Since itemServices.rentItem is expected to return a item with its availability field set to False, it might be a good idea to add in that one field to item for clarity.

synzen: oh woops good catch, you're right that I didn't actually test the renting part lol though this is the controller

https://github.com/CSC59939/QuickBorrow/blob/1c330d662ee6de6d8c81479e749ff475a5532ec9/server/controllers/items.js#L32-L39

so all it's technically doing is `res.json`ing the returned value from item services, with or without the availability field. to test the availability function you added, i would have to test the item services function that its calling (which I completely forgot to test, even though it's in the PR title lol but it's added now https://github.com/CSC59939/QuickBorrow/pull/57/commits/1c330d662ee6de6d8c81479e749ff475a5532ec9)

PR #60, opened 2019-11-11
Renting

MicLi01:
- added fields to items model for renting
- added to item services

### DIFF
--- a/server/routes/api/items.js
+++ b/server/routes/api/items.js
@@ -29,9 +29,9 @@ router.get('/:itemId', [
 
 /**
  * @memberof module:api/items
- * @name POST /rent
+ * @name GET /rent
  */
-router.get('/rent/:itemId', [
+router.get('/rent/:itemId/:borrowerId/:duration', [
   param('itemId').isAlphanumeric(),
   validatorErrors,
 ], itemController.rent);
--- a/server/routes/api/items.js
+++ b/server/routes/api/items.js
@@ -31,8 +31,10 @@ router.get('/:itemId', [
  * @memberof module:api/items
  * @name GET /rent
  */
-router.get('/rent/:itemId/:borrowerId/:duration', [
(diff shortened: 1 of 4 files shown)


server/routes/api/items.js
```diff
@@ -29,9 +29,9 @@ router.get('/:itemId', [
...
 /**
  * @memberof module:api/items
- * @name POST /rent
+ * @name GET /rent
  */
-router.get('/rent/:itemId', [
+router.get('/rent/:itemId/:borrowerId/:duration', [
   param('itemId').isAlphanumeric(),
   validatorErrors,
 ], itemController.rent);
```
synzen: this is such a security flaw lol but for now it's fine - the two extra params should be checked though
```js
param('borrowerId').isAlphanumeric(),
param('duration').isInt()
```

also preferably 
```js
/:itemId/rent/:borrowerId/:duration
```
 to keep it consistent with the other routes

server/services/items.js
```diff
@@ -14,9 +14,11 @@ const findAllItems = async () => {
   return await Item.find({});
 };
 
-const rentItem = async (id) => {
+const rentItem = async (id, borrowerId, duration) => {
   const item = await Item.findById(id);
   item.availability = false;
+  item.borrower = borrowerId;
```
synzen: the id's gotta be casted to a ObjectId

```js
const mongoose = require('mongoose');
item.borrower = new mongoose.Types.ObjectId(borrowerId);
```

server/routes/api/items.js
MicLi01: I was also planning on changing this later because revealing a user id isn't such a good idea. I made the checks for the two params and modified the route.

server/services/items.js
MicLi01: change made

PR #61, opened 2019-11-12
Fix user register route

synzen:
bunch of issues - a good indicator that we'll need tests for the users as well :>

- register route always gave an error about email exists since `User.find` always returned an array, and not `null`
- register route never sent back a response
- register route never sent back the new user in the response (this probably shouldn't be done, but for now I think it's ok since we're still developing, to be able to see the new user details)

### DIFF
--- a/server/routes/api/users.js
+++ b/server/routes/api/users.js
@@ -14,9 +14,9 @@ const {check, param} = require('express-validator');
  */
 router.post('/register', [
   check('name', 'Username field required')
-      .exists(),
+      .isLength({min: 1}),
   check('email', 'Invalid email')
-      .exists().withMessage('Email field required')
+      .isLength({min: 1}).withMessage('Email field required')
       .bail()
       .isEmail().withMessage('Not an email')
       .bail()
@@ -34,7 +34,7 @@ router.post('/register', [
  */
 router.post('/login', [
   check('email')
-      .exists().withMessage('Email field required')
+      .isLength({min: 1}).withMessage('Email field required')
       .bail()
       .isEmail().withMessage('Not an email')
       .bail()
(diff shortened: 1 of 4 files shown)


synzen:
> 
> 
> Did you happen to test for the case where a username isn't provided and when the passwords don't match? I remember testing for those cases and the errors didn't show up in the errors array when I logged it.

woops, forgot about those. those are fixed now as well

server/routes/api/users.js
```diff
@@ -14,9 +14,9 @@ const {check, param} = require('express-validator');
  */
 router.post('/register', [
   check('name', 'Username field required')
-      .exists(),
```
Akasora39: You can add a checkFalsy option to your exists call which will treat empty strings as not existing which would be the case here since the states are initialized as "".

synzen: oh good to know - though I already fixed that by replacing it with your previous call of `isLength({min: 1})` (hadn't realized `exists` didn't check the string value)

synzen: oh good to know - though I already fixed that by replacing it with your previous call of isLength({min: 1}) (hadn't realized exists didn't check the string value)

synzen: also just checked the login - for an unregistered email, it'll give 422 status code as expected
for an unknown password, it looks like it lets you login regardless - let me see

Akasora39: Yeah either way works. The checkFalsy offers more flexibility but since we're dealing with strings here it's not really necessary.

synzen:
Found out the bcrypt compare function was asynchronous - had to `await` it. everything should be gucci now

PR #66, opened 2019-11-12
Add users controller/services unit tests, bug fix

synzen:
Also

- extracted jwt sign out of controller and into services
- fixed bug for not `await`ing on bcrypt.hash since it's async

### DIFF
--- a/server/controllers/users.unit.test.js
+++ b/server/controllers/users.unit.test.js
@@ -4,7 +4,7 @@ const userServices = require('../services/users');
 jest.mock('jsonwebtoken');
 jest.mock('../services/users');
 
-describe('controller/users', function() {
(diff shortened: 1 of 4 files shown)


server/controllers/users.unit.test.js
```diff
@@ -0,0 +1,67 @@
+const userController = require('./users');
+const userServices = require('../services/users');
+
+jest.mock('jsonwebtoken');
```
JackyLiu97: I noticed that you declared `jest.mock('jsonwebtoken');` but you never used it in the unit test. Its a good idea to get rid of it

synzen: good catch, thanks

PR #67, opened 2019-11-13
Add test for Dashboard and add more tests for Register and Login

Akasora39:
Boost code coverage for frontend. More to come soon...


client/src/js/components/Login/Login.test.js
```diff
@@ -40,4 +40,34 @@ describe('Login', () => {
 		wrapper.find('input[type="password"]').simulate('change', {target: {id: 'password', value: 'password'}});
 		expect(wrapper.state('password')).toEqual('password');
 	});
+
+	test('successful login', () => {
+		let wrapper = shallow(<Login loginUser = {params.loginUser} auth = {params.auth} errors = {params.errors} />);
+		wrapper.find('input[type="email"]').simulate('change', {target: {id: 'email', value: 'test@test.com'}});
```
oscarchong: could you change all the .find(**'input'** [....) to **InputText**, since I'm using primereact for styling and this will cause error on running tests. Alongside with Register.test.js as well.

Akasora39: Gotcha. I'll make the changes after https://github.com/CSC59939/QuickBorrow/pull/68 is merged so the tests will pass on my end as well.

PR #74, opened 2019-11-13
getLendingList API

JackyLiu97:
getLendingList allows you to get all the items a user owns

### DIFF
--- a/server/routes/api/users.js
+++ b/server/routes/api/users.js
@@ -57,5 +57,14 @@ router.patch('/:userId', [
       .isAlphanumeric(),
   validationErrors,
 ], userController.edit);
+/**
+ *  @memberof module:api/users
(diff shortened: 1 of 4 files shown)


server/routes/api/users.js
```diff
@@ -57,5 +57,14 @@ router.patch('/:userId', [
       .isAlphanumeric(),
   validationErrors,
 ], userController.edit);
+/**
+ *  @memberof module:api/users
+ * @name GET /lending
+ */
+router.get('/lending/:userId/items', [
```
oscarchong: Would not recommend having "lending" before '/:userId' because this would be messy and not RESTful.

JackyLiu97: Thank you and good point. I will change that now.

PR #77, opened 2019-11-14
install react-materialize

JackyLiu97:
changes css package from primereact to react-materialize


Akasora39:
This is causing errors on the front end and the tests for it. I think it would be better to refactor the css from primereacr to materialize before merging.

oscarchong:
I would hold off on changing it because the front end for registration and login page are using primereact which caused the error on testing. Next time just make sure none of the files are using a package before deleting and changing it to another one.

PR #79, opened 2019-11-15
Add /api/items/:itemId/endrent, extract DB update logic in service to model

synzen:
- Added `/api/items/:itemId/endrent` and its respective controller and service
- Moved the database logic in `services/items` as a method of Item in `/models/Item`
- Updated tests

let me know what you guys think about this change in logic

(i seem to have a tendency to constantly refactor lol)

### DIFF
--- a/server/models/Item.js
+++ b/server/models/Item.js
@@ -29,7 +29,7 @@ async function rentTo(borrowerId, duration) {
  */
 async function endRent() {
   this.set({
-    availability: false,
+    availability: true,
     borrower: undefined,
   });
   await this.save();
--- a/server/models/Item.js
+++ b/server/models/Item.js
@@ -28,7 +28,6 @@ async function rentTo(borrowerId, duration) {
  * Stop renting an item
  */
 async function endRent() {
-  this.availability = true;
   this.set({
     availability: false,
     borrower: undefined,
--- a/server/models/Item.js
+++ b/server/models/Item.js
@@ -11,5 +11,35 @@ const ItemSchema = new Schema({
   returnDate: {type: Date, default: Date.now},
 });
 
+/**
+ * Rent an item to a borrower for a specified duration
+ * @param {mongoose.Types.ObjectId|string} borrowerId - The borrow's ObjectId
+ * @param {number|string} duration - ms since January 1, 1970, 00:00:00 UTC
+ */
+async function rentTo(borrowerId, duration) {
+  this.availability = false;
+  this.borrower = new mongoose.Types.ObjectId(borrowerId);
+  this.returnDate = new Date((new Date()).getTime() + Number(duration));
+  await this.save();
+  return this;
+}
+
+/**
+ * Stop renting an item
+ */
+async function endRent() {
+  this.availability = true;
+  this.set({
+    availability: false,
+    borrower: undefined,
+  });
+  await this.save();
+  return this;
+};
+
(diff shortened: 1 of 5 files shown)


server/models/Item.js
```diff
@@ -11,5 +11,34 @@ const ItemSchema = new Schema({
...
+ * Stop renting an item
+ */
+async function endRent() {
+  this.set({
+    availability: false,
+    borrower: undefined,
+  });
+  await this.save();
+  return this;
+};
```
MicLi01: availability here should be set to true for ending rent.

synzen: good catch, fixed ~~

synzen:
> 
> 
> availability should be true for ending rent. It would probably be a good idea to save the transaction so a user can view their past transactions.

that's a good point. for that I'm thinking of a transaction model

| Column  | Type | Purpose |
| ------------- | ------------- | -- |
| id  | ObjectId  | |
| itemId  | ObjectId | |
| borrower | ObjectId | |
| start | Date | Start of trans |
| end | Date | End of trans |
| token? | String | Maybe the token thing yu feng was talking about? this could probably be the `id` |


this can remove the `availability` since that can be derived from the start/end date, and the `borrower` as well

Akasora39:
> > availability should be true for ending rent. It would probably be a good idea to save the transaction so a user can view their past transactions.
> 
> fixed the availability, thanks
> 
> that's a good point. for that I'm thinking of a transaction model
> 
> Column	Type	Purpose
> _id	ObjectId	Built in I'd
> itemId	ObjectId	
> borrower	ObjectId	
> start	Date	Start of trans
> end	Date	End of trans
> token?	String	Maybe the token thing yu feng was talking about? this could probably be the `_id` tbh
> this can remove the `availability` from the `Item` since that can be derived from the transaction start/end dates, and the `borrower` as well

We were discussing 

> > availability should be true for ending rent. It would probably be a good idea to save the transaction so a user can view their past transactions.
> 
> fixed the availability, thanks
> 
> that's a good point. for that I'm thinking of a transaction model
> 
> Column	Type	Purpose
> _id	ObjectId	Built in I'd
> itemId	ObjectId	
> borrower	ObjectId	
> start	Date	Start of trans
> end	Date	End of trans
> token?	String	Maybe the token thing yu feng was talking about? this could probably be the `_id` tbh
> this can remove the `availability` from the `Item` since that can be derived from the transaction start/end dates, and the `borrower` as well

We had some further discussion and the token may not be necessary as we can essentially get all the information we need from the itemId and borrower.

yliu0715:
> > availability should be true for ending rent. It would probably be a good idea to save the transaction so a user can view their past transactions.
> 
> fixed the availability, thanks
> 
> that's a good point. for that I'm thinking of a transaction model
> 
> Column	Type	Purpose
> _id	ObjectId	Built in I'd
> itemId	ObjectId	
> borrower	ObjectId	
> start	Date	Start of trans
> end	Date	End of trans
> token?	String	Maybe the token thing yu feng was talking about? this could probably be the `_id` tbh
> this can remove the `availability` from the `Item` since that can be derived from the transaction start/end dates, and the `borrower` as well

Let's not worry about the token thing for now. It is still a bit fuzzy in my head as to how we are going to have confirmation on borrowing and renting. We can leave that for the last releases if we have time.

synzen:
sounds guud

PR #81, opened 2019-11-15
Add validators for routes with object ids

synzen:
The server currently throws a 500 Internal Server Error if you pass an invalid mongoDB id to one of our routes that requires it - this will check before it reaches the error handler, and return a properly formed error


synzen:
> 
> 
> Nice catch. That might be something we want to include in our tests.

indeed, forgot about that - will be in a future PR

PR #84, opened 2019-11-16
Oc/change to material ui

oscarchong:
- changed package from 'primereact' to 'material ui'

- changed all related files that's using primereact to material ui

- fixed all unit tests to pass on material ui stuff

PS: sorry this PR is kind of too big


client/src/js/components/Login/Login.js
```diff
@@ -62,11 +63,12 @@ export class Login extends Component {
...
 		{
-			return "";
+			return true;
 		}
 		else
 		{
-			return errors.errors.find(x => x.param === e).msg;
+			// return errors.errors.find(x => x.param === e).msg;
+			return false;
 		}
```
synzen: since true/false is being returned here to match Material UI's `TextField` error property requirements, we'll have to have another way to display the error in a future PR

PR #85, opened 2019-11-16
Frontend tests

Akasora39:
- Added index.js and serviceWorker.js to be ignored when testing for coverage
- Fixed loginUser action test

### DIFF
--- a/client/src/js/actions/index.test.js
+++ b/client/src/js/actions/index.test.js
@@ -1,5 +1,11 @@
 import axios from 'axios';
 import * as actions from './index';
+import setAuthToken from "../utils/setAuthToken";
+import jwt_decode from "jwt-decode";
+
+jest.mock('axios')
+jest.mock('../utils/setAuthToken')
+jest.mock('jwt-decode')
 
 describe('actions', () => {
 	let mock;
@@ -22,20 +28,30 @@ describe('actions', () => {
 	    expect(history.push).toHaveBeenCalledWith('/login');
 	});
 
-	test('loginUser', async () => {
-	    const dispatch = jest.fn();
+	describe('loginUser', () => {
+		const dispatch = jest.fn();
 	    const userData = {
 	    	email: 'test@test.com',
 	    	password: 'password'
 	    }
-	    const mockedError = {
-	    	response: {
-	    		data: 'test error'
-	    	}
-	    }
-	    mock.mockResolvedValue().mockRejectedValueOnce(mockedError);  // mock axios.post to resolve
-	    await actions.loginUser(userData)(dispatch);
-	    expect(mock).toHaveBeenCalledWith('/api/users/login', userData);  // Success!
+		test('successful post made', async () => {
+			mock.mockResolvedValue({data: {}});
+		    await actions.loginUser(userData)(dispatch);
+		    expect(mock).toHaveBeenCalledWith('/api/users/login', userData); 
+		});
+		test('error thrown', async () => {
+			const mockedError = {
+		    	response: {
+		    		data: 'test error'
+		    	}
+		    }
+		    mock.mockResolvedValue().mockRejectedValueOnce(mockedError);
+		    await actions.loginUser(userData)(dispatch);
+		    expect(mock).toHaveBeenCalledWith('/api/users/login', userData); 
+		});
+	    
+	    
+	    
 
 	});
 
--- a/client/src/js/actions/index.test.js
+++ b/client/src/js/actions/index.test.js
@@ -1,11 +1,6 @@
 import axios from 'axios';
 import * as actions from './index';
 
-const mockResponse = {
-	success: true,
-	token: "Bearer eyJhbGciOiJIUzI.eyJpZCI6IjVjN.1Tn8FLJEGGE8"
-};
-
 describe('actions', () => {
 	let mock;
 	beforeEach(() => {
--- a/client/src/js/actions/index.test.js
+++ b/client/src/js/actions/index.test.js
@@ -1,7 +1,6 @@
 import axios from 'axios';
 import * as actions from './index';
-import setAuthToken from "../utils/setAuthToken";
-import jwt_decode from "jwt-decode";
+import { GET_ERRORS } from "./types";
 
 jest.mock('axios')
 jest.mock('../utils/setAuthToken')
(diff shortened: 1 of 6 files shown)


client/src/js/actions/index.test.js
```diff
@@ -22,16 +22,22 @@ describe('actions', () => {
...
 	    	email: 'test@test.com',
 	    	password: 'password'
 	    }
-	    mock.mockResolvedValue();  // mock axios.post to resolve
+	    const mockedError = {
+	    	response: {
+	    		data: 'test error'
+	    	}
+	    }
+	    mock.mockResolvedValue().mockRejectedValueOnce(mockedError);  // mock axios.post to resolve
```
synzen: was this mock rejected value added by mistake?

Akasora39: No for some reason this mock in particular would throw an error so I had to mock the error to cover that part. Still not sure why it doesn't process the request successfully but at the very least this test covers the error handling.

Akasora39:
> you really want that coverage lol

Lol from what I've seen serviceWorker and index.js files are typically ignored. Maybe we should be more flexible with the index.js file since that holds some functionality with regards to the verification token but for now I think we can ignore it.

client/src/js/actions/index.test.js
synzen: you passed nothing to the `mockResolvedValue()` so in this snippet in index.js:
https://github.com/CSC59939/QuikBorrow/blob/9a14a996aa4effa36ee3a55928dca82de7c0f2ec/client/src/js/actions/index.js#L40-L51

`res` is undefined, so you're trying to access data of undefined. so then it throws a TypeError, and proceeds to the catch:
https://github.com/CSC59939/QuikBorrow/blob/9a14a996aa4effa36ee3a55928dca82de7c0f2ec/client/src/js/actions/index.js#L52-L57
which is then another error because `err` is undefined, so you have to mock the rejected value for that bug lol

do 
```js
mock.mockResolvedValueOnce({data: {}})
```
and it should fix it - the empty data object is also necessary since you're destructuring `res.data`

synzen: i would also suggest doing

```js
jest.mock('axios')
jest.mock('../utils/setAuthToken')
jest.mock('jwt-decode')
```

if this is a unit test to make sure all external calls are not actually being called

Akasora39: Ah I see. I was trying to create a mock response for that and tried to call a mock store but had some issues with that. Who knew the solution would be as simple as that lol. Expected it to work without issues since I based it off the register test but the register test wasn't doing anything with the response data so I can see why there weren't any issues there. Thanks!

yliu0715:
@Akasora39 I was reading up on notifications and it has a lot to do with serviceWorkers. So we might need it when we set up push notifications. Do you think this might be a hassle in the future if you ignore it right now?

client/src/js/actions/index.test.js
```diff
@@ -22,16 +28,32 @@ describe('actions', () => {
...
+		test('error thrown', async () => {
+			const mockedError = {
+		    	response: {
+		    		data: 'test error'
+		    	}
+		    }
+		    mock.mockResolvedValue().mockRejectedValueOnce(mockedError);
+		    await actions.loginUser(userData)(dispatch);
+		    expect(mock).toHaveBeenCalledWith('/api/users/login', userData); 
+		});
```
synzen: looks like you're making the same assertions in both tests - I think a better test for "error thrown" would be to assert that dispatch was called if the fetch failed

```js
import { GET_ERRORS } from "./types";

test('error dispatched', async () => {
  const dispatch = jest.fn()
  const mockedError = {
      response: {
        data: 'test error'
      }
    }
  mock.mockRejectedValueOnce(mockedError);
  await actions.loginUser(userData)(dispatch);
  expect(dispatch).toHaveBeenCalledWith({
    type: GET_ERRORS,
    payload: mockedError.response.data
  })
});
```
(also to ensure that the same dispatch function is not shared between multiple tests)

client/src/js/actions/index.test.js
```diff
@@ -1,5 +1,11 @@
 import axios from 'axios';
 import * as actions from './index';
+import setAuthToken from "../utils/setAuthToken";
+import jwt_decode from "jwt-decode";
```
synzen: the imports can be left out btw since `jest.mock` works on the node require (process-wide) cache, so it'll be mocked when it's called in other files within the same process too

client/src/js/actions/index.test.js
synzen: no problemo

client/src/js/actions/index.test.js
Akasora39: Yeah I see. I seem to have a habit of using assertions as a way of expecting no error to occur in the process of the test but that's not a good solution.

client/src/js/actions/index.test.js
Akasora39: Interesting. I'll keep this in mind for the future. Was planning on using those to expect that the functions were called but it seems that those lines were covered anyways.

Akasora39:
> @Akasora39 I was reading up on notifications and it has a lot to do with serviceWorkers. So we might need it when we set up push notifications. Do you think this might be a hassle in the future if you ignore it right now?

No, I initially ignored the serviceWorker file because I didn't really understand what it was doing that impacted the crux of our system. If you plan on using notifications through the serviceWorker, then we will need to test it in the future. For now since we aren't really using it, I believe it can be ignored in regards to code coverage but that doesn't really change much. All ignoring does will not count it towards the code coverage report generated and we can change this in the future when a test is written for the serviceWorker.js file.

client/src/js/components/Login/Login.test.js
```diff
@@ -57,17 +57,20 @@ describe('Login', () => {
...
 
 	test('errors', () => {
 		const fakeErr = {
 			errors: [{
-				'email': 'Invalid email'
+				'param': 'email',
+				'msg' : 'Invalid email'
 			}]
 		}
 		let wrapper = shallow(<Login loginUser = {params.loginUser} auth = {params.auth} errors = {params.errors} />);
```
yliu0715: Instead of defining `let wrapper = shallow(<Login loginUser = {params.loginUser} auth = {params.auth} errors = {params.errors} />);` in all your test functions, you can probably abstract it out by declaring it in a beforeEach like this:

`beforeEach(() => {
  let wrapper = shallow(<Login loginUser = {params.loginUser} auth = {params.auth} errors = {params.errors} />);
})`

Akasora39: Good point. I figured there would be more instances where I had to pass specific props into the wrapper that would make each rendering different enough to warrant separate definitions but it seems that isn't the case.

PR #87, opened 2019-11-18
added transaction

MicLi01:
Created transactions. Notifications can can be done using this by getting all the unprocessed transactions for a specific user.

### DIFF
--- a/server/routes/api/transactions.js
+++ b/server/routes/api/transactions.js
@@ -2,10 +2,7 @@
 
 const express = require('express');
 const router = new express.Router();
-const {param} = require('express-validator');
 const transactionController = require('../../controllers/transactions');
-const validatorErrors = require('../../middleware/shared/validatorErrors');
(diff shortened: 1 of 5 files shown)


server/routes/api/transactions.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+router.post('/approve', transactionController.approve);
+
+/**
+ * @memberof module:api/transactions
+ * @name POST /reject
+ */
+
+router.post('/reject', transactionController.reject);
```
synzen: the `transactionId` should ideally be a parameter instead of in the body to match our other routes

`POST /api/transactions/:transactionId/approve`
`POST /api/transactions/:transactionId/reject`

but tis a minor thing. the middleware validators seem are also missing to check for valid inputs before handing it off to the controller. these can be left to a future PR though I guess, if you want

MicLi01: will keep that in mind when I add in some additional things for notifications

PR #89, opened 2019-11-18
Migrate NavBar to Material UI, remove Materialize

synzen:
- Added the missing Roboto font (https://material-ui.com/components/typography/)
- Removed materialize 
- Updated navbar to Material UI

image

our color scheme (and material UI's default color scheme) is quite cancerous if I do say so myself - we might want to adjust it https://material-ui.com/customization/theming/

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -24,7 +24,6 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">
     <title>React App</title>
   </head>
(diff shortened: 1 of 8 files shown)


client/public/index.html
```diff
@@ -24,7 +24,6 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">
     <title>React App</title>
```
oscarchong: This isn't necessary, but could you change "React App" to "QuikBorrow" since you're editing this file? Other than that, everything else is perfect.

synzen: _anything_ for you hun 👍

PR #90, opened 2019-11-18
added transaction lists

MicLi01:
Originally I wasn't going to have a lender field because I thought I could get that from item but then I would have needed to check each item to check if it was for that user. I was only able to do this by getting all the unprocessed request first which could end up using up a lot of time. So i decided to add in a lender field to Transaction.

### DIFF
--- a/server/services/transactions.js
+++ b/server/services/transactions.js
@@ -18,9 +18,29 @@ const rejectTransaction = async (id) => {
   return updated;
 };
 
+const getUserBorrowTransactions = async (userId) => {
+  const transactions = await Transaction.find({borrower: userId});
+  return transactions;
+};
+
+const getUserPendingTransactions = async (userId) => {
+  const transactions = await Transaction
+      .find({borrower: userId, processed: false});
+  return transactions;
+};
+
+const getTransactionRequests = async (userId) => {
+  const requests = await Transaction
+      .find({lender: userId, processed: false});
+  return requests;
+};
+
 module.exports = {
   createTransaction,
   approveTransaction,
   rejectTransaction,
+  getUserBorrowTransactions,
+  getUserPendingTransactions,
+  getTransactionRequests,
 };
 
--- a/server/services/transactions.js
+++ b/server/services/transactions.js
@@ -18,29 +18,23 @@ const rejectTransaction = async (id) => {
   return updated;
 };
 
-const getUserBorrowTransactions = async (userId) => {
-  const transactions = await Transaction.find({borrower: userId});
+const getTransactions = async (userId, type, isProcessed) => {
+  let query;
+  console.log(type);
+  if (type == 'borrower') {
+    query = {borrower: userId};
+  } else {
+    query = {lender: userId};
+  }
+  query['processed'] = isProcessed;
+  const transactions = await Transaction.find(query);
   return transactions;
-};
-
-const getUserPendingTransactions = async (userId) => {
-  const transactions = await Transaction
-      .find({borrower: userId, processed: false});
-  return transactions;
-};
-
-const getTransactionRequests = async (userId) => {
-  const requests = await Transaction
-      .find({lender: userId, processed: false});
-  return requests;
 };
 
 module.exports = {
   createTransaction,
   approveTransaction,
   rejectTransaction,
-  getUserBorrowTransactions,
-  getUserPendingTransactions,
-  getTransactionRequests,
+  getTransactions,
 };
 
--- a/server/controllers/transactions.js
+++ b/server/controllers/transactions.js
@@ -33,29 +33,12 @@ const reject = async (req, res, next) => {
       .catch(next);
 };
 
-const getUserBorrowTransactions = async (req, res, next) => {
-  transactionServices.getUserBorrowTransactions(req.body.userId)
+const getTransactions = async (req, res, next) => {
+  transactionServices.getTransactions(req.body.userId,
+      req.query.type, req.query.isProcessed)
       .then((transactionList) => {
         console.log(transactionList);
         res.json(transactionList);
-      })
-      .catch(next);
-};
-
-const getUserPendingTransactions = async (req, res, next) => {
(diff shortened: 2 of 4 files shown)


MicLi01:
Also made the route changes that Michael T suggested on previous pull request.

synzen:
the routes are getting quite complicated lol - I think the ideal way to do this would be to use [queries](https://expressjs.com/en/api.html#req.query), and build on top of the users API for the PR's goals, for example:

`GET /api/users/:userId/transactions` - gets all transactions of all types for this user, where they're either the lender or borrower

`GET /api/users/:userId/transactions?type=borrower` - gets all transactions of all types for this user, where this user is the borrower

`GET /api/users:/:userId/transactions?status=pending` - gets all transactions where they're unprocessed for this user

`GET /api/users/:user/transactions?type=borrower&status=approved` - gets all transactions where they're approved, and he's the borrower for this user

this _greatly_ reduces code duplication with sensible API routes, because otherwise we need to specify a new route for each and every one of these cases lol. but there might be too much rewriting from what you have now so I guess it's up to you - getting it up and running might be a higher priority

MicLi01:
> 
> 
> the routes are getting quite complicated lol - I think the ideal way to do this would be to use [queries](https://expressjs.com/en/api.html#req.query), and build on top of the users API for the PR's goals, for example:
> 
> ```js
> // api/users
> router.get('/:userId/transactions, userController.getTransaction);
> ```
> 
> `GET /api/users/:userId/transactions` - gets all transactions of all types for this user, where they're either the lender or borrower
> 
> `GET /api/users/:userId/transactions?type=borrower` - gets all transactions of all types for this user, where this user is the borrower
> 
> `GET /api/users:/:userId/transactions?status=pending` - gets all transactions where they're unprocessed for this user
> 
> `GET /api/users/:user/transactions?type=borrower&status=approved` - gets all transactions where they're approved, and he's the borrower for this user
> 
> this _greatly_ reduces code duplication, requiring only one sensible API route, because otherwise we need to specify a new route for each and every one of these cases lol. but there might be too much rewriting from what you have now so I guess it's up to you - getting it up and running might be a higher priority

Yea I also felt the code duplication was quite high. I was aiming to get things up and running first and then refactor later on.

synzen:
> 
> Yea I also felt the code duplication was quite high. I was aiming to get things up and running first and then refactor later on.

that will also mean that the tests we write for this are gonna have to be delayed until the rewrite, idk how @Akasora39 feels about that lol

honestly I can do the rewrite and testing for it later if not anyone else, if @Akasora39 's okay with it

MicLi01:
> 
> 
> > Yea I also felt the code duplication was quite high. I was aiming to get things up and running first and then refactor later on.
> 
> that will also mean that the tests we write for this are gonna have to be delayed until the rewrite, idk how @Akasora39 feels about that lol
> 
> honestly I can do the rewrite and testing for it later if not anyone else, not a problem for me, if @Akasora39 's okay with it (our situation here is a beautiful thing called tech debt lol)

oh thats true. I forgot to consider the tests. I can refactor this sometime tomorrow then.

Akasora39:
> > Yea I also felt the code duplication was quite high. I was aiming to get things up and running first and then refactor later on.
> 
> that will also mean that the tests we write for this are gonna have to be delayed until the rewrite, idk how @Akasora39 feels about that lol
> 
> honestly I can do the rewrite and testing for it later if not anyone else, not a problem for me, if @Akasora39 's okay with it (our situation here is a beautiful thing called tech debt lol)

No worries lol. I just think we should get in the habit of including tests with our prs so it'll come to us naturally as we make more prs. In the first iteration, test driven development seemed harder since we were mostly following tutorials but at this point I think we're comfortable enough with the codebase to consider it.

synzen:
sounds guud fellas

MicLi01:
> > > Yea I also felt the code duplication was quite high. I was aiming to get things up and running first and then refactor later on.
> > 
> > 
> > that will also mean that the tests we write for this are gonna have to be delayed until the rewrite, idk how @Akasora39 feels about that lol
> > honestly I can do the rewrite and testing for it later if not anyone else, not a problem for me, if @Akasora39 's okay with it (our situation here is a beautiful thing called tech debt lol)
> 
> No worries lol. I just think we should get in the habit of including tests with our prs so it'll come to us naturally as we make more prs. In the first iteration, test driven development seemed harder since we were mostly following tutorials but at this point I think we're comfortable enough with the codebase to consider it.

I will add in the tests in this pull request.

MicLi01:
Actually, I will add in the tests in a separate pull request after speaking with Andy about them tomorrow.

server/services/transactions.js
```diff
@@ -18,9 +18,23 @@ const rejectTransaction = async (id) => {
...
+  console.log(type);
+  if (type == 'borrower') {
+    query = {borrower: userId};
+  } else {
+    query = {lender: userId};
+  }
+  query['processed'] = isProcessed;
+  const transactions = await Transaction.find(query);
+  return transactions;
+};
```
synzen: there's currently no way here to get transactions of all types since it defaults to lender if there is no type so i would suggest:

```js
const getTransactions = async (userId, type, isProcessed) => {
  let query = {};
  if (type === 'borrower') {
    query.borrower = userId;
  } else if (type === 'lender') {
    query.lender = userId;
  }

  if (isProcessed === 'true' || isProcessed === 'false') { // strings because req.query are always passed as strings
    query.processed = isProcessed;
  }

  const transactions = await Transaction.find(query);
  return transactions;
};
```

MicLi01: When I was doing this, I was thinking that a user wouldn't need to ever get all transactions at the same time since the list would be cluttered. I was thinking that the front end would have some kind of selectors for borrower/lender and processed/unprocessed to get a focused list of things they want. I can modify it if you feel it could be useful.

synzen: ah gotcha, I just thought that if the user did happen to need all their transactions (for example to display on their profile or whatever), we would be forced to make multiple requests on the front end. but it's fine, we're not there yet :P

PR #92, opened 2019-11-19
added user profile reducers in redux,  added testing for user profile reducer and action

JackyLiu97:


client/src/js/reducers/userReducer.js
```diff
@@ -0,0 +1,16 @@
+import { GET_USER_PROFILE, GET_ERRORS } from '../actions/types';
+
+const initState = {};
+
+const userReducer = (state = initState, action) => {
+    switch (action.type) {
+      case GET_USER_PROFILE:
+        return action.payload;
+      case GET_ERRORS:
+          return action.payload;
```
yliu0715: Minor change but you can reduce those two case statements to one line:
`case GET_USER_PROFILE: case GET_ERRORS:
        return action.payload;`

client/src/js/actions/types.js
```diff
@@ -3,3 +3,4 @@ export const FETCH_PRODUCT = "fetch_product";
 export const GET_ERRORS = "GET_ERRORS";
 export const USER_LOADING = "USER_LOADING";
 export const SET_CURRENT_USER = "SET_CURRENT_USER";
+export const GET_USER_PROFILE = "USER_PROFILE";
```
yliu0715: should probably assign it `"GET_USER_PROFILE"` for consistency.

client/src/js/reducers/userReducer.js
JackyLiu97: ah I see. I think i will keep it as it is for completeness <3

client/src/js/actions/types.js
JackyLiu97: Good idea! consistency is good <3

oscarchong:
Please assign reviewers to your PR or else we might over look it leading to no reviews on it because we might think it's not ready for review or something. Thank you.

JackyLiu97:
@synzen  kek.

PR #95, opened 2019-11-19
Setup search feature on backend

Akasora39:
Updated items to include route for searching. Ideally, the URL will look something like ...com/search?param=somequery
Based on this, I plan on having a search bar in the navbar and a separate component for search results. This will likely be what I work on for the next pr.

Test Coverage:
Frontend: N/A
Backend: +0%

### DIFF
--- a/server/routes/api/items.js
+++ b/server/routes/api/items.js
@@ -48,4 +48,10 @@ router.get('/:itemId/endrent', [
   validatorErrors,
 ], itemController.endRent);
 
+/**
+ * @memberof module:api/items
+ * @name POST /search
+ */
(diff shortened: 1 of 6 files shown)


server/routes/api/items.js
```diff
@@ -48,4 +48,10 @@ router.get('/:itemId/endrent', [
   validatorErrors,
 ], itemController.endRent);
 
+/**
+ * @memberof module:api/items
+ * @name POST /search
+ */
+router.get('/search', itemController.search);
+
```
MicLi01: minor thing here. description says post but router uses get.

PR #99, opened 2019-11-19
Add basic modal design for adding items with their images

synzen:
image

for this PR it's all purely visual atm (except the file drag n drop, that's actually working) - you can see it if you click the button on the homepage

test coverage +0.37


client/src/js/components/AddItemModal/AddItemModal.test.js
```diff
@@ -0,0 +1,41 @@
...
+import {act} from 'react-dom/test-utils'
+import AddItemModal from './AddItemModal';
+import Enzyme, {mount} from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('AddItemModal', () => {
+	it('renders', () => {
+        const wrapper = mount(<AddItemModal open={true} onClose={jest.fn()} />)
```
Akasora39: Just wanted to note that for instances where you'll be simulating a component a lot, it'd make your code cleaner to have a beforeEach that will call the mount.

client/src/js/components/AddItemModal/AddItemModal.test.js
```diff
@@ -0,0 +1,41 @@
...
+		const imageDropzone = wrapper.find('div[data-testid="image-dropzone-container"]')
+		const images = [ new File([], 'image1.jpg'), new File([], 'aaaa.png'), new File([], 'aedgtrhrf.jpg') ]
+		act(() => {
+			imageDropzone.prop('onDrop')(images)
+		})
+		wrapper.update()
+		wrapper.find(`*[data-testid="delete-image-${images[0].name}"]`).hostNodes().simulate('click')
+		wrapper.find(`*[data-testid="delete-image-${images[2].name}"]`).hostNodes().simulate('click')
+		expect(wrapper.find(`*[data-testid^="delete-image"]`).hostNodes()).toHaveLength(1)
+		expect(wrapper.find(`*[data-testid="delete-image-${images[1].name}"]`).hostNodes()).toHaveLength(1)
```
Akasora39: So if I'm understanding this correctly, it seems like you are deleting the first and third image attached and then checking that there should only be one image remaining and it should be the second one attached. This makes sense but I feel something like checking that the click action deleted the image successfully would be a more straightforward approach.

synzen: oh yea I get what you mean as it seems like a roundabout way of accomplishing the test. my logic was basically researching react testing (https://reactjs.org/docs/testing.html), which then led me to their recommended tool (https://testing-library.com/docs/react-testing-library/intro), which then led me to their core principle: "The more your tests resemble the way your software is used, the more confidence they can give you."

it made a lot of sense to me - so in this case I tested what the user would see after clicking the two delete buttons (they should only see 1 element, and that element should the 2nd image). I avoided testing the implementation details (the internal state, I'm guessing is what you meant?) because it may lead to brittle tests (tests breaking if we change the implementation details of how images are removed for example, which would then require refactoring tests every time we make such changes)

I'm still new to testing react so lmk what you think :P

client/src/js/components/AddItemModal/AddItemModal.test.js
synzen: sounds gooood, done

client/src/js/components/AddItemModal/AddItemModal.test.js
Akasora39: Yeah I get what you mean. I think just having it check that the removeFile function is called would suffice since that should remain the same regardless of how we choose to remove images. This situation seems a bit overboard for removing a file since testing the removal of one file would indicate that removing multiple files also works.

synzen: ohh, maybe you misunderstood my purpose of this test (the test name is probably misleading) - this test in particular is for the view (what the user sees), not the "business" logic behind it. basically pretend I'm a user and do what they would do. for this, I had to find the relevant nodes in the render tree to assert their existence (hence why it looks a bit complicated, the attributes are a bit long lol)

though your last last assumption (removing one file working assumes that removing multiple will work the same) I'll have to disagree there :P for example, if I passed in the same index 0 to the removeFile function every time by mistake, and then I try to delete the third item, then the first item would be removed instead causing the current test to rightfully fail, but if I just asserted the removeFile function was called, the test would pass and it wouldn't give insight as to why it's not working as expected

but btw the inner (private) functions of a Hook, which is what removeFile is, are not exposed since they're defined inside. so I couldn't assert on it unless I convert it back to class syntax. But having it like this happens to fall in line with React's recommended testing approach of "[test React components without relying on their implementation details](https://reactjs.org/docs/testing.html)" (but devil's advocate, it's recommended, not mandatory xP)

synzen: I _can_ try to add another test for the implementation detail (asserting private function calls and whatnot), it is a valid approach of white box testing, but I'll probably have to revert back to class syntax

Akasora39: I see your point. Perhaps a better test would be to expect that the right file is removed so maybe something like checking that the image is there before deleting, then simulating the delete action, and expecting what was deleted to not be present anymore. Although this already sounds rather complex.

I think my issue with the test was that it didn’t directly suggest that the first and third image were deleted because it only checked that the second was still there.

synzen: ahh ok I see, I made a slight change in the test now - check it out

no need to explicitly check if the nodes on there beforehand in this case because if they aren't, then `.simulate` will throw an error causing the test to fail if it didn't find any

Akasora39: Ah ok. Alright looks good!

PR #105, opened 2019-11-21
Add and fix most server code documentation

synzen:
- Disabled the default jsdoc rules provided by eslint since they're [deprecated](https://eslint.org/blog/2018/11/jsdoc-end-of-life), so I added the eslint plugin [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc) rules to handle code documentation instead
- Fixed and added much documentation to many backend functions following these new rules

(if you use vscode and don't have the eslint plugin installed, it's probably a good idea to do so :>)

i'm guessing @Akasora39 is the documenter - if Nikolai ever questions your role, just give him this PR lol

### DIFF
--- a/server/.eslintrc.js
+++ b/server/.eslintrc.js
@@ -5,6 +5,7 @@ module.exports = {
   },
   'extends': [
     'google',
+    'plugin:jsdoc/recommended'
   ],
   'globals': {
     'Atomics': 'readonly',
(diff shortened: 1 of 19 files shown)


synzen:
aha you guys right

but dw guys, plenty of documentation could be written on the client code :>

PR #106, opened 2019-11-21
Add POST /api/users/:userId/items with auth

synzen:
this route will allow a new item to be created for a certain user, with authentication, requiring incoming requests to have a [Authorization header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) with the JWT token

I also made a tiny refactor in documenting - the repeated import doc line
```js
/** @type {import('express').RequestHandler} */
```
can be reduced to just
```js
/** @type {RequestHandler} */
```
if we include a `@typedef` at the top for `RequestHandler`

I made that mistake in some other files too like server/controllers/items in #105 , if you guys wanna _nudge nudge_ get some _nudge nudge_ documentation in _wink wink_

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -1,4 +1,5 @@
 const userServices = require('../services/users');
+const itemServices = require('../services/items');
 
 // eslint-disable-next-line valid-jsdoc
 /** @type {import('express').RequestHandler} */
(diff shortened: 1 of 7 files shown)


synzen:
> 
> 
> Everything you do is beautiful

PR #107, opened 2019-11-21
Add DELETE /api/users/:userId/items/:itemId with auth

synzen:
this route will allow an item to be deleted, verified by authentication via JWT token provided in [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) header and item owner's user ID compared to the id in the decoded JWT token and the userId param

I also think sending errors back as arrays is a bit problematic for sending back API errors - lmk if you guys have any ideas

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -90,10 +90,10 @@ const deleteItem = async (req, res, next) => {
     });
   }
 
-  /** @type {import('mongoose').Document} */
+  /** @type {RequestHandler} */
   const item = req.item;
   if (item.get('user', String) !== userId) {
-    return res.status(401).json({
+    return res.status(404).json({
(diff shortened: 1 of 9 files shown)


server/controllers/users.js
```diff
@@ -78,11 +78,35 @@ const createItem = async (req, res, next) => {
...
+    });
+  }
+
+  /** @type {RequestHandler} */
+  const item = req.item;
+  if (item.get('user', String) !== userId) {
+    return res.status(404).json({
+      errors: [{msg: 'Unauthorized (user does not own this item)'}],
+    });
+  };
```
yliu0715: Looks good, but wouldn't it make more sense to put these 3 checks in the middleware folder?

synzen: it has been done, my liege!

PR #108, opened 2019-11-22
Add search frontend

Akasora39:
Setup search action, reducer, and component. 
The Search component is located in the Products page and will send a query to the items and return documents matching the query by name. 
Search acts as a filtering of sorts and won't refresh upon refreshing the page i.e. if you search for a chair and try to refresh the products page, it will still only display a chair product.


client/src/js/components/Search/Search.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { searchProducts } from '../../actions';
+import styled from 'styled-components';
```
oscarchong: minor thing, but since you're not using styled, you can just remove it.

client/src/js/components/Products/ProductList.js
```diff
@@ -107,6 +108,7 @@ class ProductList extends Component {
           <List>
             {items}
           </List>
+          <Search />
```
yliu0715: Doesn't this put the search component at the bottom?

Akasora39: Yeah, there's an issue right now with the Navbar which hides things at the top like the pagination so I kept it at the bottom for now.

client/src/js/actions/index.js
```diff
@@ -20,6 +20,26 @@ export const fetchProduct = itemId => {
...
 };
 
+/**
+ *  Search by query for items in database by calling {@link Action} to backend
+ *  @param {String} query - query string to search
+ *  @param {Object} function - used to dispatch actions
+ *  @return {Promise<Object>} - response from an {@link Action}
+ */
+export const searchProducts = query => dispatch => {
+  return axios
```
synzen: so a mistake here in these docs here - `searchProducts` is a function that returns a function (with dispatch as a param) that returns an axios request (rather than, what your doc says, a function with a param named "function" that returns a promise)

you can check for correctness if you try to type `dispatch` to see if Intellisense in VS Code can detect the var type (`any` means no type specified)
image



an interesting problem - I did some searching and found https://github.com/jsdoc/jsdoc/issues/1286. one of the solutions mentioned there is putting another function doc right before dispatch: 

image

and now VS code can see the `dispatch` data type, and is not `any` anymore. but I think there are other, cleaner ways to do it written on that issue so other ways could work. also TIL they're called "curried" functions

synzen: as a side note, it looks putting the search results in the global redux state may be going overboard since there's only 1 component that needs it afaik, and that's the product list lol

Akasora39: :O Did not realize that. I'll have to get something similar for the text editor I'm using to verify this. For now, I'll document it like this before testing other ways.

synzen: don't tell me you're using Notepad! lol

Akasora39:
> Also, i think putting search results in the global redux state might be a little too much since only the product list component needs search results

For now that may be true but I can see search being utilized for other components as well depending on how it's abstracted. Right now it only searches based on name but this can be changed to search based on other attributes such as owner which can then be used to display products a user is lending.

oscarchong:
> > Also, i think putting search results in the global redux state might be a little too much since only the product list component needs search results
> 
> For now that may be true but I can see search being utilized for other components as well depending on how it's abstracted. Right now it only searches based on name but this can be changed to search based on other attributes such as owner which can then be used to display products a user is lending.

thats no problem if you can see it being utilize for other components in the future!

PR #110, opened 2019-11-23
quick fix on margins

oscarchong:
Navbar was covering logo on login, register page, and others, this is a fix to all of it

Screen Shot 2019-11-22 at 9 02 48 PM
Screen Shot 2019-11-22 at 9 17 28 PM


oscarchong:
> Niceu, I think the same needs to be done for Register and ProductList too

will fix that right now!

Issue #102, opened 2019-11-20
Image (Storing and retrieving)


yliu0715:

JackyLiu97:
Is there any resources for storing and retrieving an image in react?

PR #111, opened 2019-11-24
Add actual request to add item in add item modal

synzen:
testing react is the bane of my existence

- Added login check
image
- Added loading state (while it's fetching, inputs disabled):
image
- Added errors:
image
- Added button to go to product page on success
image


synzen:
related to #102 

images are ready to be sent to the server with the commented code here:

https://github.com/CSC59939/QuikBorrow/blob/691e2b3f7c00825d5b21532e68283be5c8ee7d7f/client/src/js/components/AddItemModal/AddItemModal.js#L129-L134

But using the commented code requires the backend route to handle `FormData`, which is different from now since we're only handling JSON and `req.body` will be empty with `FormData`. 

To handle files from the client, we'll need a [package to handle it](https://stackoverflow.com/questions/24800511/express-js-form-data), and the one I recommend is [express-formidable](https://www.npmjs.com/package/express-formidable) since it's the easiest to use

I'll uncomment that part once the server route's been updated to handle store images in this route

yliu0715:
> related to #102
> 
> images are ready to be sent to the server with the commented code here:
> 
> https://github.com/CSC59939/QuikBorrow/blob/691e2b3f7c00825d5b21532e68283be5c8ee7d7f/client/src/js/components/AddItemModal/AddItemModal.js#L129-L134
> 
> But using the commented code requires the backend route to handle `FormData`, which is different from now since we're only handling JSON and `req.body` will be empty with `FormData`.
> 
> To handle files from the client, we'll need a [package to handle it](https://stackoverflow.com/questions/24800511/express-js-form-data), and the one I recommend is [express-formidable](https://www.npmjs.com/package/express-formidable) since it's the easiest to use
> 
> I'll uncomment that part once the server route's been updated to handle store images in this route

@synzen To handle form data, I'm using a package called Multer. It's basically a middleware that handles form data which is mainly used for uploading files like images. Right now I'm storing the images we received from the frontend in a static directory called `uploads` on our server. But ultimately, this will be a heavy load on the server when we get a shit ton of images and I think we might want to store and serve through Firebase which will act as a CDN and we can store the path to the image on Mongo. I haven't looked into express-formidable, but do you know the tradeoffs between the two (Multer and express-formidable)? If express-formidable does a better job for us, then I can refactor the server to use it. Also, if we were to convert to using Firebase, do you think we should do the storing to Firebase on the client side or server side?

synzen:
> 
> @synzen To handle form data, I'm using a package called Multer. It's basically a middleware that handles form data which is mainly used for uploading files like images. Right now I'm storing the images we received from the frontend in a static directory called `uploads` on our server. But ultimately, this will be a heavy load on the server when we get a shit ton of images and I think we might want to store and serve through Firebase which will act as a CDN and we can store the path to the image on Mongo. I haven't looked into express-formidable, but do you know the tradeoffs between the two (Multer and express-formidable)? If express-formidable does a better job for us, then I can refactor the server to use it. Also, if we were to convert to using Firebase, do you think we should do the storing to Firebase on the client side or server side?

yea I was looking at multer since it is one of the packages mentioned in the link I posted - what I didn't like about it was its verbosity:

#### [multer](https://www.npmjs.com/package/multer#readme)
```js
app.post('/profile', upload.single('avatar'), function (req, res, next) {
  // req.file is the `avatar` file
  // req.body will hold the text fields, if there were any
})
app.post('/photos/upload', upload.array('photos', 12), function (req, res, next) {
  // req.files is array of `photos` files
  // req.body will contain the text fields, if there were any
})
```
#### [express-formidable](https://www.npmjs.com/package/express-formidable)
```js
// index.js
app.use(formidableMiddleware());
```
```js
// routes
app.post('/upload', (req, res) => {
  req.fields; // contains non-file fields
  req.files; // contains files
});
```
but either is fine I guess, it's not like we have 20 routes to handle images :P 

about using Firebase, I never used it - maybe it'll introduce unnecessary overhead since it's an external service and it's not like we're going international (yet lol) with a heavy load. I was initially thinking of just storing the images as base64 strings (images can be represented as strings), or apparently as [Buffers](https://stackoverflow.com/questions/44869479/what-data-type-should-i-use-to-store-an-image-with-mongodb) since it looks like mongodb supports it. _but_ if you've already got it figured out though, by all means use Firebase lol

Akasora39:
I noticed that you used the act function from react-dom/test-utils instead of the one from the react-test-renderer package that was already installed. Was there a reason for that?

synzen:
> 
> 
> I noticed that you used the act function from react-dom/test-utils instead of the one from the react-test-renderer package that was already installed. Was there a reason for that?

oh I didn't even notice - at the time I wrote it, no reason since I just let VS Code auto-import it for me lol but according to the react docs they function similarly, maybe the one from `react-test-renderer` is specifically for renders made by `react-test-renderer`. tbh the act there feels like a hacky solution but it was the only way for me to get it working with Enzyme without errors

Issue #114, opened 2019-11-24
Filters for ProductList via Transactions


synzen:
Depending on whether they've been lended out or not. The current /api/transactions route does not accommodate this feature since this is specific to _items_ not a user

there also seems to be another problem where we don't know how long the lend is, so adding a `start` date and `end` date to Transaction might be beneficial

synzen:
if someone wants to do this, they can unassign me and assign themselves for them extra commits lol

Issue #112, opened 2019-11-24
Borrow a product on ProductList


synzen:
This one's a bit hard

1. Make sure they're eligible to borrow (client-side)
    - They don't own the item
    - They're not currently borrowing it
    - The item is available for rent (others are not borrowing it)
2. Make the request to the server
    - Make the same checks as above
3. Update the products list

PR #115, opened 2019-11-24
Fix log in persistence

synzen:
this will fix it so Andy's(?) code in index.js actually works and keeps us logged in

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -8,7 +8,7 @@ import reduxThunk from "redux-thunk";
 import jwt_decode from "jwt-decode";
 import setAuthToken from "./js/utils/setAuthToken";
 import { setCurrentUser, logoutUser } from "./js/actions/index";
-
+import { Provider } from "react-redux";
 import reducers from "./js/reducers";
 
 const store = createStore(reducers, {}, applyMiddleware(reduxThunk));
@@ -33,7 +33,9 @@ if (localStorage.jwtToken) {
 }
 
 ReactDOM.render(
-    <App />,
+	<Provider store={store}>
+    	<App />
+	</Provider>,
   document.getElementById("root")
 );
 
--- a/client/src/js/actions/index.js
+++ b/client/src/js/actions/index.js
@@ -46,6 +46,7 @@ export const loginUser = userData => dispatch => {
       setAuthToken(token);
       // Decode token to get user data
       const decoded = jwt_decode(token);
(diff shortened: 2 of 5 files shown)


synzen:
> 
> 
> It looks like it fixed the log in persistence problem and when I do a full refresh now, it doesn't log me out. Now when the app starts up I start off logged in so that is an issue.

yeah that's the (Andy's(?)) intended behavior from the code - when the app starts up, or when it changes that forces a refresh, that's the same as a manual page refresh so a refresh so it shouldn't be logging the user out - only when they _explicitly_ click log out is their token in `localStorage` deleted, and they're require to log in again

the jwt token is set in `localStorage` on login:
https://github.com/CSC59939/QuikBorrow/blob/64527cb98c27be6cf963042cd0cb299e4dc878e7/client/src/js/actions/index.js#L37-L44

and whenever the page loads the javascript, andy's(?) code here runs

https://github.com/CSC59939/QuikBorrow/blob/64527cb98c27be6cf963042cd0cb299e4dc878e7/client/src/index.js#L16-L33

and the `localStorage.jwtToken` should always be defined because the only place where it is deleted is when they manually log out here:

https://github.com/CSC59939/QuikBorrow/blob/64527cb98c27be6cf963042cd0cb299e4dc878e7/client/src/js/actions/index.js#L75-L78

synzen:
yeah those still need to be fixed - could open an issue and assign to whoever wants it ^^

synzen:
> 
> 
> I was going to fix this after I could finally merge my search frontend pr but looks like you beat me to it lol. I was initially just going to move the code from the index to the App.js file since I believe the original issue was moving the store to get the test for the Navbar to pass. But it looks like you're solution is simpler and much cleaner! No complaints from me.

o yea it wasn't working before cause it was using the `store.dispatch` of a store that wasn't actually handed off to a Provider to the App, so it was a "dangling store" with no use

https://github.com/CSC59939/QuikBorrow/blob/12f69887657bcac409a8170228370ae2d1c26662/client/src/index.js#L14-L38

so I just moved it back to its original place

https://github.com/CSC59939/QuikBorrow/blob/64527cb98c27be6cf963042cd0cb299e4dc878e7/client/src/index.js#L35-L40

updated tests, and vOiLa! also yeah, I started to work on #112 , but then that needed me to work on #114 , but then trying to do that I was so frustrated by the login issue I had to fix this lmao

PR #116, opened 2019-11-25
reject all other transaction requests on approval and modified tests

MicLi01:
cancels other transactions on the same item on approval of item.
modified tests to reflect this.

### DIFF
--- a/server/models/Transaction.int.test.js
+++ b/server/models/Transaction.int.test.js
@@ -23,19 +23,6 @@ describe('Transaction Model Test', () => {
     });
   });
 
-  it('create & save transaction successfully', async () => {
-    const validTransaction = new TransactionModel(transactionData);
-    const savedTransaction = await validTransaction.save();
-    // Object Id should be defined when successfully saved to MongoDB.
-    expect(savedTransaction._id).toBeDefined();
-    expect(savedTransaction.borrower).toBeDefined();
-    expect(savedTransaction.lender).toBeDefined();
-    expect(savedTransaction.msg).toBe(transactionData.msg);
-    expect(savedTransaction.duration).toBe(transactionData.duration);
-    expect(savedTransaction.item).toBeDefined();
-    expect(savedTransaction.approved).toBe(transactionData.approved);
-    expect(savedTransaction.processed).toBe(transactionData.processed);
-  });
   it('approves a transaction successfully', async () => {
     const validTransaction = new TransactionModel(transactionData);
     const savedTransaction = await validTransaction.approve();
--- a/server/models/Transaction.int.test.js
+++ b/server/models/Transaction.int.test.js
@@ -1,0 +1,49 @@
+const mongoose = require('mongoose');
+const TransactionModel = require('./Transaction');
+const transactionData = {borrower: '5db9fdce1dd490177413ab0b', lender: '5dc4967ed2616e19c8eef654', msg: 'test transaction', duration: 1, item: '5dc9dba21c9d4400001b972d', approved: false, processed: false};
+const keys = require('../config/keys');
+const MONGODB_OPTIONS = {
+  useNewUrlParser: true,
+  useCreateIndex: true,
+};
+
+describe('Transaction Model Test', () => {
+  beforeAll(async () => {
+    await mongoose.connect(keys.mongoURI, MONGODB_OPTIONS, (err) => {
+      if (err) {
+        throw err;
+      }
+    });
+  });
+
+  it('create & save transaction successfully', async () => {
+    const validTransaction = new TransactionModel(transactionData);
+    const savedTransaction = await validTransaction.save();
+    // Object Id should be defined when successfully saved to MongoDB.
+    expect(savedTransaction._id).toBeDefined();
+    expect(savedTransaction.borrower).toBeDefined();
+    expect(savedTransaction.lender).toBeDefined();
+    expect(savedTransaction.msg).toBe(transactionData.msg);
+    expect(savedTransaction.duration).toBe(transactionData.duration);
+    expect(savedTransaction.item).toBeDefined();
+    expect(savedTransaction.approved).toBe(transactionData.approved);
+    expect(savedTransaction.processed).toBe(transactionData.processed);
+  });
+  it('approves a transaction successfully', async () => {
+    const validTransaction = new TransactionModel(transactionData);
+    const savedTransaction = await validTransaction.approve();
+    expect(savedTransaction.approved).toBe(true);
+    expect(savedTransaction.processed).toBe(true);
+  });
+  it('rejectss a transaction successfully', async () => {
+    const validTransaction = new TransactionModel(transactionData);
+    const savedTransaction = await validTransaction.reject();
+    expect(savedTransaction.approved).toBe(false);
+    expect(savedTransaction.processed).toBe(true);
+  });
+});
+
+afterAll( async () =>{
+  await mongoose.connection.close();
+});
+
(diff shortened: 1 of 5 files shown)


server/services/transactions.unit.test.js
```diff
@@ -27,8 +29,12 @@ describe('services/item', function() {
     it('approves a transaction by id', async function() {
       const id = 'abc123';
       Transaction.findById.mockResolvedValueOnce({approve: jest.fn()});
+      Transaction.updateMany.mockResolvedValueOnce({});
+      itemServices.rentItem.mockResolvedValueOnce({'duration': 98});
       await transactionServices.approveTransaction(id);
       expect(Transaction.findById).toHaveBeenCalledWith(id);
+      expect(Transaction.updateMany).toHaveBeenCalled();
+      expect(itemServices.rentItem).toHaveBeenCalled();
```
Akasora39: Is there a reason for mocking the resolved value for rentItem? It doesn't seem like you're doing anything with it in particular but maybe I'm missing something that's going on between the item and transaction.

server/services/transactions.js
```diff
@@ -24,7 +24,16 @@ const createTransaction = async (data) => {
  */
 const approveTransaction = async (id) => {
   const transaction = await Transaction.findById(id);
+  await Transaction.updateMany(
+      {'item': transaction.item},
+      {'$set': {'processed': true}},
+  );
```
Akasora39: I was a bit confused at first because of the updateMany call but essentially what you're doing here is finding the item identified by id and updating it to setting processed to true. Is there a reason for using updateMany vs updateOne? For updating an item, there should be a unique id for each so there wouldn't be a case where more than 1 item has the same id so I'm not sure why updateMany was used.

MicLi01: I think you understood it wrong. I'm not finding the item. I'm finding all the open transactions having to do with that item and setting processed to true since I am approving one of the transactions here which means it will be unavailable.

server/services/transactions.unit.test.js
MicLi01: Since this is unit test for transactions I am abstracting away functionality having to do with items.

server/services/transactions.js
Akasora39: Ah that makes more sense. Might want to include that in part of the documentation.

server/services/transactions.unit.test.js
Akasora39: Right but what is the explicit purpose of mocking a resolved here? `itemServices.rentItem.mockResolvedValueOnce({'duration': 98});` 
It doesn't seem like you're doing anything with the value returned.

synzen:
btw I noticed the approval route should also check if the item is available before renting it out, and if not return a 403 (kinda off topic from this PR, but I guess you can include it in this one if you want)

server/services/transactions.js
MicLi01: Yea I created a few items to see if it processed would be set to true for the other ones and it worked.

MicLi01:
> 
> 
> btw I noticed the approval route should also check if the item is available before renting it out, and if not return a 403 (kinda off topic from this PR, but I guess you can include it in this one if you want)
> 
> also do you know yet how the `duration` you added here is going to interact with an Item's `availability`?

Yea I'll add in a check for item availability, that seems useful. Duration would just tell when the borrower has to return it by. In an earlier pull request I made something that stores the return by date in an item's field. The borrower will have something on the front end to end the rent and if the current date is past the due date, then they will be charged a fee. I noticed that we also don't have a price in item so that will need to be added in.

synzen:
hmm ok I see. I feel that `availability` and definitely the `returnDate` relates more to transactions than they do to the actual item model itself, but it seems to work atm so ¯\_(ツ)_/¯

also there's the problem of how the server is going to know when the due dates in the DB has passed (maybe via polling), but that's a problem for another day when we get borrowing to work within the next century

MicLi01:
> hmm ok I see. I feel that `availability` and definitely the `returnDate` relates more to transactions than they do to the actual item model itself, but it seems to work atm so ¯_(ツ)_/¯
> 
> also there's the problem of how the server is going to know when the due dates in the DB has passed (maybe via polling), but that's a problem for another day when we get borrowing to work within the next century

Now that I am thinking about it, do we need a rent route? The way I'm doing it now makes it so when a transaction is approved, the item will be rented out so I don't think we would need a rent route and we can get rid of it.

server/services/transactions.unit.test.js
MicLi01: @Akasora39 you are right. I don't need to mock the values here since I didn't do anything with them(I thought I did).

synzen:
good point, that looks like a no to me

server/models/Transaction.int.test.js
```diff
@@ -0,0 +1,56 @@
...
+    // Object Id should be defined when successfully saved to MongoDB.
+    expect(savedTransaction._id).toBeDefined();
+    expect(savedTransaction.borrower).toBeDefined();
+    expect(savedTransaction.lender).toBeDefined();
+    expect(savedTransaction.msg).toBe(transactionData.msg);
+    expect(savedTransaction.duration).toBe(transactionData.duration);
+    expect(savedTransaction.item).toBeDefined();
+    expect(savedTransaction.approved).toBe(transactionData.approved);
+    expect(savedTransaction.processed).toBe(transactionData.processed);
+  });
```
synzen: this particular test is not necessary since we're not responsible for testing other module's implementations (`.save`) - in this case, mongodb and mongoose's teams are supposed to write their tests to assert that this function works, not us - but the rest of the integration tests below this looks good.

server/models/Transaction.int.test.js
```diff
@@ -0,0 +1,56 @@
...
+const MONGODB_OPTIONS = {
+  useNewUrlParser: true,
+  useCreateIndex: true,
+};
+
+describe('Transaction Model Test', () => {
+  beforeAll(async () => {
+    await mongoose.connect(keys.mongoURI, MONGODB_OPTIONS, (err) => {
+      if (err) {
+        throw err;
```
synzen: also I'm not sure if @Akasora39  ever set it up so we're not using our actual production database collection for these tests (otherwise we're filling our db with garbage), but this looks like we are using our production db

server/models/Transaction.int.test.js
MicLi01: For this one I was testing more so that the database connection string that I used was correct first.

server/models/Transaction.int.test.js
MicLi01: And for this one theres no problem because if you don't specify a dbName in the MONGODB_Options it defaults to the tests database which is where everything here is getting saved to.

server/models/Transaction.int.test.js
MicLi01: I thought it would be good to know if the connection was up first.

synzen: oh i see, tho tbh that makes this test case a bit more odd/unrelated considering the test name & assertions made lol but it looks like the db connection is already tested within your `beforeAll`, where it should throw the error in the callback if mongoose experienced an error connecting:

https://github.com/CSC59939/QuikBorrow/blob/5a28a910a81b89b06ea3ab7e210f195dabccb660/server/models/Transaction.int.test.js#L18-L24

if there is no error thrown in there we should be able to assume that it was a successful connection. 

but btw by default, model queries through mongoose like `.save` doesn't throw an error if there is no connection (see https://mongoosejs.com/docs/connections.html#buffering) - in reality the promise would never resolve because it's basically queued up, which will then fail the test with time-out errors (something like "test did not finish within expected timeframe"), which can ultimately mislead debugging efforts if this behavior is not immediately known. I think it (like many other things in mongoose) is truly awful default behavior but ¯\_(ツ)_/¯

server/models/Transaction.int.test.js
synzen: oh gotcha

server/models/Transaction.int.test.js
MicLi01: I didn't consider the before all so I think its best to remove this test.

PR #117, opened 2019-11-25
Fix user auth error displays and changed search

Akasora39:
Fixed error displaying on register and login. Due to changes of switching to material-ui, the TextField would require a boolean value for the errors field. Thus, the getErrors function had to be changed to return boolean values instead. Fix reverted to the original getErrors function and checked whether or not the corresponding error was empty (==="").

Changes to search:

- Moved search to be above the products and minor style changes
- Added a button to correspond to the search submit event
- Limited character entry for query to prevent bad requests
- Entering search query now redirects you to the first page of the products page. Prior to this, if you were on the 5th page and entered a query that would result in 1 item being displayed, you would still be on the 5th page with nothing shown until you manually navigated back to the 1st page through the pagination.

### DIFF
--- a/client/src/js/components/Login/Login.js
+++ b/client/src/js/components/Login/Login.js
@@ -63,12 +63,11 @@ export class Login extends Component {
 		const { errors } = this.state;
 		if (errors.errors === undefined || errors.errors.find(x => x.param === e) === undefined)
 		{
-			return true;
+			return "";
 		}
 		else
 		{
-			// return errors.errors.find(x => x.param === e).msg;
-			return false;
+			return errors.errors.find(x => x.param === e).msg;
 		}
 	};
 	render() {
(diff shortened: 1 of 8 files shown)


client/src/js/components/Login/Login.js
```diff
@@ -83,7 +82,7 @@ export class Login extends Component {
 						<TextField
 							onChange={this.onChange}
 							value={this.state.email}
-							error={this.getErrors('email')}
+							error={this.getErrors('email')===""}
```
oscarchong: Here you're not showing the actual error text, cause error is getting a boolean value, you should add another prop named `defaultValue = {this.getErrors('email')}` to show the error text. It would be the same case for password also and registration page. You can see more info on material UI's documentation 
https://material-ui.com/components/text-fields/
and this ss is more specific on what I was mentioning. 
Screen Shot 2019-11-25 at 11 27 16 PM

Akasora39: Actually, the error text is being displayed through the spans below. That was how it initially displayed errors and seeing how it was left there, I thought it'd be okay to continue doing so. I can change it so it's displayed through the actual TextField component though.

PR #120, opened 2019-11-27
Yl/images

yliu0715:
- added Multer middleware for handling multipart/form-data from client
- created a static directory `uploads` which stores the images received
- added a environment variable `DOMAIN` which serves the purpose of creating a path for the uploads directory
- stored a pointer (path) to the image in `uploads` on MongoDB
- Added new rule for eslint to handle linebreak issues


server/middleware/items.js
```diff
@@ -1,5 +1,25 @@
...
+  filename: function(req, file, cb) {
+    cb(null, new Date().toISOString().replace(/:/g, '-') + file.originalname);
+  },
+});
+
+const imageFilter = (req, file, next) => {
+  if (!file.originalname.match(/\.(jpg|jpeg|png|gif)$/)) {
+    return next(new Error('Only image files are allowed!'), false);
+  }
+  next(null, true);
```
Akasora39: Seems like lines 7, 10, 15, 16, and 18 aren't covered by the current tests. Can you add some more tests for the new functionality?

MicLi01:
Adding in the images for products is good. Profiles will also need a image too so it can make a account more personalized. Think we can use a similar format to this to handle that. Using that way we can get rid of most of our placeholder images.

client/src/js/components/Homepage/Homepage.js
```diff
@@ -78,7 +79,9 @@ export function Homepage() {
 						I want to lend
 					</Button>
 					<Button size='large' variant='contained'>
-						I want to borrow
+						<Link to='/products' style={{textDecoration: 'none', color: 'black'}}>
+							I want to borrow
+						</Link>
 					</Button>
```
synzen: to keep the UI experience consistent, it's better to just programmatically route the user to the /products page instead. right now the button is partially broken:
image
clicking the padding (red circles) around the text will do nothing, since the link is around the text and not the button. to programmatically route via history in hooks, you can just first import

```js
import { useHistory } from 'react-router-dom';
```
and then in the hook itself:
```js
const history = useHistory()
// ...
<Button size='large' variant='contained' onClick={() => history.push('/products')}>
I want to borrow
</Button>
```

ezpz, praise hooks

server/middleware/items.js
yliu0715: @Akasora39 I have addressed lines 14 to 18 by writing new tests. I'm not sure if there is a way to test the diskStorage function. Still need more time to look into that.

client/src/js/components/Homepage/Homepage.js
yliu0715: @synzen Yes, React Hooks is definitely something I need to pick up.

yliu0715: @synzen Can you take a look at the merge conflict on package-lock.json. I don't want to accidentally delete something we need.

synzen: all fixed - it was just that you added a package that master's package-lock didn't have, but for some reason it was still a merge conflict

yliu0715: I have addressed some of the changes you mentioned: giving the button an `onClick` property that routes the user to the `/products` page using the history in hooks and renaming the `productImage` field to `imagePath`. As for the inconsistency in naming, I honestly had no idea how that happened loool, but I agree with you that it will take too much resources to make that change now given the limited time we have.

PR #123, opened 2019-11-30
Revert "Oc/profile frontend"

Akasora39:
Test files should not be removed outright. I'm not sure what the issue was on your end but from my testing on Wednesday, it seemed like there was no need to adjust the tests too much since it was mostly changes to the frontend which is covered by a render test. In the future, when an issue comes up with your tests, please discuss it with others first to see if you can resolve it.


MicLi01:
I agree with this. We should be trying to commit code with tests instead of removing them.

oscarchong:
Yes, will be commit code with test in the future. sorry bout that.

PR #96, opened 2019-11-19
Oc/profile frontend

oscarchong:
- Redesign of the profile page (Improved UI color + design to have consistency in frontend design)
- lending list is now updated with what user uploaded
- Replaced mock info with user information 

Screen Shot 2019-11-26 at 6 45 02 PM

PS: tests are removed, will rewrite for next PR.

PR #124, opened 2019-12-01
Wrote test for profile files PR #96

oscarchong:
I made a mistake of removing test PR #96. So this PR are the tests for it and it include Profile.test.js and ProfileCard.test.js. Apologize for removing the test on that PR, I will write test for the files I've edited or create and push it in the same PR in the future. 

PS: first time writing tests and thanks for Michael Tan and Andy for the guidance


client/src/js/components/Profile/Profile.test.js
```diff
@@ -0,0 +1,21 @@
...
+        const user ={}
+        const match = { params: { profileId: 'foo' } }
+		let component = renderer.create(
+        <UserProfile 
+        getUserProfile={() => {}}
+        fetchProducts={() => {}}
+        match = {match}
+        user = {user}
+        products = {[]}
+        />);
```
Akasora39: An id is not being passed here for the user so 
`if (!_id) {
            return <h3> User profile does not exist</h3>
        }
        else {
            return (
            <ProfileCard
            name={name}
            age={age}
            college={college}
            products={lendinglist}
            bio={bio}
            wishlist={wishlist}
            rating={rating}
            email={email}
            />
                )` only the case where there's no such user profile is being tested. Can you include a test so that a valid id is provided?

client/src/js/components/Profile/ProfileCard.test.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import ProfileCard from './ProfileCard';
+
+
+describe('ProfileCard', () => {
```
Akasora39: The products list and wishlist list that was recently added in PR #96 is not covered either.  Here is something I found that might apply to what you have for the ProfileCard https://stackoverflow.com/questions/40600706/testing-jsx-inside-map-function-using-enzyme

oscarchong: thanks for the reference, I will write more tests to cover the missing parts

client/src/js/components/Profile/ProfileCard.test.js
```diff
@@ -0,0 +1,34 @@
...
+    const profile = mount(
+        <ProfileCard
+        products ={[]}
+        wishlist = {[]}
+        />);
+
+    it('render list', () =>{
+        expect(profile.find('#products-list').children()).toEqual(expect.anything());
+        expect(profile.find('#wish-list').children()).toEqual(expect.anything());
+	});
```
Akasora39: It seems like here you are asserting that the products list and wish list are not null which intuitively makes sense. However, I think it would be better to just check that they each have a length of 1. Otherwise, this test could suggest that you could have more than 1 product list or wish list and still pass.

oscarchong: I think i was just not clear when i say "render list", this is more "render list items" like checking if there's any items in the product list and wishlist and it could be any number.

Akasora39: Oh wait I forgot that this was checking the list elements itself and not the actual list. Although that does raise the point that you should check the product list itself and not the actual children since it's possible to have a list that's empty.

oscarchong: With way ProfileCard is written, it’s not possible to test products_li cause its a local var
Screen Shot 2019-12-07 at 1 06 16 AM

Akasora39: Really? I was able to get it working. I did `expect(profile.find('#products-list').hostNodes()).toHaveLength(1);` and the tests were passing just fine. Either way, I think checking the children of the list is a bad call since they are meant to be of variable length so we shouldn't expect them to have a specific length.

oscarchong: You're right, i just tried and it works, i was trying yesterday and it failed for some reason but it might be some typo or something, but i did a quick fix. Should be good now. Thanks a lot

PR #125, opened 2019-12-02
Add Codecov to CI for automatic test coverage reports on PRs

synzen:
And voila, we have it working for both server and client files! Codecov will merge the two coverage reports into one big one for an overall coverage report. I think we can also set target coverage numbers as well, but lmk what you guys think

Any further commits after the initial report will cause the bot to edit their report too

*Praise Github Actions!*


synzen:
we can also split it up even further to have codecov calculate coverage *per directory* (like controller coverage, services coverage, models coverage, etc.), but that might be too much

Akasora39:
I like what I'm seeing! I agree that checking each individual directory may be a bit overkill. I remember you mentioned how there may be an issue with checking frontend and backend individually. Is that still an issue or could we just check the paths for 'client/' and 'server/'? Also, there were some files that were ignored in Jest due to irrelevance with the crux of our application so I'm assuming we would have to manually configure that for codecov as well.

synzen:
> 
> 
> I like what I'm seeing! I agree that checking each individual directory may be a bit overkill. I remember you mentioned how there may be an issue with checking frontend and backend individually. Is that still an issue or could we just check the paths for 'client/' and 'server/'? Also, there were some files that were ignored in Jest due to irrelevance with the crux of our application so I'm assuming we would have to manually configure that for codecov as well.

not an issue :D I managed to figure it out, it's like I mentioned in the first post, codecov is able to merge the coverage reports from both frontend back backend and combine the percentages together (like how you see codecov's post in this PR)


so how it works is (for example, for the frontend), whenever github runs these tests on their servers as part of our CI:
1. they generate the test coverage report file/artifacts(?) with the `npm run test-client-coverage` script I added with the jest config `"coverageReporters": ["lcov"],` in package.json (you can see this in the changed files)
2. codecov finds the relevant files (the `lcov` file) and downloads it to their servers via Github Actions
3. the same thing is done for the backend. `npm run test-server-coverage`, but in a separate environment (part of Github Actions Node CI / Server Tests environment. step 1 is in the Node CI / Client Tests environment)
4. codecov intelligently waits for both to be downloaded and analyzed, and finally creates one ultimate

step 4 here was critical to making this work, and it works beautifully

oscarchong:
This is a good idea and it would be good to set a test coverage target.

PR #127, opened 2019-12-04
Rewrite error reducer to accomodate all errors

synzen:
accommodating all types of errors will be needed for handling so many errors that originate from fetches for borrowing, lending, login, register, getting user profile, etc.

I organized the actions like here:

image

and most of the changes are just renaming. with this, the naming convention is very important since that's how the error reducer distinguishes what actions it should be handling


synzen:
> 
> 
> i would recommend clearing the errors whenever FETCH or FINISHED action happens so a component doesn't believe that there's an error by mistake when it is fetching something or when it already fetched something.

good catch. will do - this also made me realize that the errors should probably be cleared every time the route changes as well. will edit to do this too.

synzen:
actually, if we want to implement a clear errors action that happens every time the route changes, it's probably best to leave that in another PR to keep PR sizes manageable

client/src/js/reducers/errorReducer.js
```diff
@@ -1,12 +1,50 @@
...
+   * by creating a clone to maintain immutability
+   */
+  if (isFinish) {
+    return Object.keys(state).reduce((newState, key) => {
+      if (key !== stateKey) {
+        newState[key] = state[key]
+      }
+      return newState
+    }, {})
+  }
```
yliu0715: Looks good, but one thing that confused me is why do we need this specific function here? If we encounter an action of type FINISHED which means that the request is made and we successfully got a response, couldn't we just clear the error reducer? which you have already done on lines 17-19. Maybe I'm not getting the definition of FINISHED and ERROR? Hmm

synzen: ah late reply sorry, gertner has me slaving away overtime - that's because each fetch action has its own error object within the error state (which is the whole point of this PR), for example
```js
// error state
{
  GET_PROFILE_ERROR: {},
  UPDATE_PROFILE_ERROR: {}
}
```
and etc. Like 17-19 refers to an action that will clear *all* errors, but otherwise a `_FINISH` action for a particular fetch will only remove its respective error object within the error state, which is the `stateKey`

the snippet you  referred to is making a new object, and copying all the error objects from the original state, *except* the `stateKey` from the `_FINISH` action as a way to remove its respective error object while keeping the original object unmutated

synzen:
hey guys, it's been 5 days lol

PR #129, opened 2019-12-05
Comments backend

Akasora39:
Setup backend for comments


MicLi01:
I feel that including the rating with the item could be useful.

MicLi01:
> 
> 
> I feel that including the rating with the item could be useful.

Oh my mistake I meant with the comment

Akasora39:
> adding in rating into comments? feels like it might belong with a comment because its similar to a review.

I also thought about that but items already have a rating system. We should be consistent with one way to rate because allowing ratings with and without comments seems weird.

Akasora39:
> You should merge master for the updated codecov checks and it's good to go

Not sure that's necessary since the coverage reflected previously still matches my recent changes and I had no issues with codecov.

oscarchong:
> > You should merge master for the updated codecov checks and it's good to go
> 
> Not sure that's necessary since the coverage reflected previously still matches my recent changes and I had no issues with codecov.

alright just wanted to stay consistent with the "checks" on PR since there's only 4 checks here, but its not a problem

Annotation 2019-12-07 202711

PR #131, opened 2019-12-07
adding tests for items, removing unneeded route, adding fields to item

MicLi01:
deleted the renting route. renting will be done when transaction is approved now.
missing tests in items models from first sprint
added price and rating to item model


MicLi01:
rating could be taken out of items model since the comments model that andy made can hold a rating field so the item rating could be determined from the average of those.

server/controllers/items.js
```diff
@@ -6,6 +6,7 @@ const create = async (req, res, next) => {
     name: req.body.name,
     description: req.body.description,
     image: req.body.image,
+    price: req.body.price,
```
yliu0715: @MicLi01 Btw, the actual route that we are using to create an item is in `/controllers/users.js`, so I feel like the `price` field should be there instead.

server/models/Item.int.test.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+describe('Item Model Test', () => {
+  beforeAll(async () => {
+    await mongoose.connect(keys.mongoURI, MONGODB_OPTIONS, (err) => {
+      if (err) {
+        throw err;
+      }
+    });
+  });
```
yliu0715: @MicLi01 Probably should use a test database here so we don't populate our actual db.

server/controllers/items.js
MicLi01: oh i see the route will make the change

server/models/Item.int.test.js
MicLi01: Michael T commented the same thing on my previous pull request. If theres no dbname specified like above then its written to the test database by default.

server/services/transactions.js
```diff
@@ -54,9 +54,13 @@ const rejectTransaction = async (id) => {
...
- * - userId
- * - type
- * - processed
+- userId
+- type
+- processed
+ *
+ * @param userId
+ * @param type
+ * @param isProcessed
```
synzen: the comments are invalid (as jdoc says) - the format should be
`@param {type} variableName`

so in this case

```js
@param {string} userId
@param {string} type
@param {boolean} isProcessed
```
if you use VSCode it should tell you if you're writing it correctly if it provides autocomplete

server/routes/api/items.js
```diff
@@ -40,19 +40,6 @@ router.get('/:itemId', [
...
- * @memberof module:api/items
- * @name GET /rent
- */
-router.get('/:itemId/rent/:borrowerId/:duration', [
-  param('itemId').custom(isObjectId),
-  param('borrowerId').custom(isObjectId),
-  param('duration').isInt(),
-  validatorErrors,
-], itemController.rent);
-
```
synzen: if we're deleting this then we should delete the relevant controller, services and the model methods too no?

MicLi01: Still need the item services because i use it in transactions. Ill delete the rent from controller.

server/services/transactions.js
MicLi01: Good to know. I'll make the change.

PR #133, opened 2019-12-08
Add more specific availability text to items

synzen:
In both the product list and the product page, it will show one of three cases
- they own the item
- the item's being lended out
- they're currently borrowing the item
- they're eligible to borrow it

this will allow the user to see that they are actually borrowing the item, after the (to-be-implemented) borrow action is run, part of the process for #112 

image
image


client/src/js/components/Products/ProductCard.js
```diff
@@ -53,21 +56,30 @@ const FlexCardContent = styled(CardContent)`
...
+    : !availability
+      ? <span style={{ color: grey[500] }}>Currently being lended out</span>
+      : borrower === loggedInUser.id
+        ? <span style={{ color: green[500] }}>You're currently borrowing this item!</span>
+        : <span>Available to borrow</span>
 
   return (
     <FlexCard>
-      <FlexCardImage image={image || 'https://patch.com/img/cdn/users/1142384/2013/09/raw/77d3e8242e7562885116ebff68689271.jpg'} title='image' />
+      <FlexCardImage image={'https://patch.com/img/cdn/users/1142384/2013/09/raw/77d3e8242e7562885116ebff68689271.jpg'} title='image' />
```
yliu0715: @synzen We don't want to render the image on the ProductCard no more? :(

synzen: nah all our products are peanut butter jelly times now

kidding lol, this was the old Buffer-type image that we're not using anymore but I'll add it back to avoid conflicts

synzen:
Ignore codecov CI failure for this. for some reason, it's only comparing the coverage to only the server coverage and not client/server coverage combined

synzen:
> 
> 
> This is nice. Items looks much better with this. I remember seeing that we were still displayed the userId on the items recently. Was that ever taken out or modified? If not, its something to consider.

i replaced the user id with this since the user themselves wouldn't have a reason to see a user's id, but showing the user's name on the product page might be another feature

PR #144, opened 2019-12-11
Separate actions for users to own file

Akasora39:
As pointed out in https://github.com/CSC59939/QuikBorrow/pull/118, the index.js file for actions was getting cluttered with various actions focused on different components. 
This pr moves the `registerUser, loginUser, setCurrentUser, setUserLoading, and logoutUser` actions to its own users.js file. 
All imports that used any of these actions have been adjusted accordingly. 
I wasn't sure if the getUserProfile should go with it since these actions were mostly for user authentication.

### DIFF
--- a/client/src/js/components/Login/Login.js
+++ b/client/src/js/components/Login/Login.js
@@ -2,7 +2,7 @@ import React, { Component } from "react";
 import { Link } from "react-router-dom";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
-import { loginUser } from "../../actions/index";
+import { loginUser } from "../../actions/users";
 import classnames from "classnames";
 import styled from 'styled-components';
 import quikLogo from '../../components/quikLogo.png';
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -7,7 +7,7 @@ import { createStore, applyMiddleware } from "redux";
 import reduxThunk from "redux-thunk";
 import jwt_decode from "jwt-decode";
 import setAuthToken from "./js/utils/setAuthToken";
-import { setCurrentUser, logoutUser } from "./js/actions/index";
+import { setCurrentUser, logoutUser } from "./js/actions/users";
 import { Provider } from "react-redux";
 import reducers from "./js/reducers";
 import { BrowserRouter } from "react-router-dom";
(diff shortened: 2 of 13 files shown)


Akasora39:
> I think getUserProfile should go into "actions/users.js" as well just because it's user based and having all "users" in the same file would be more organized and "actions/index.js" can be all "products" stuff.

Alright. It would probably be more cluttered when there are actions added for editing the profile but we don't have that yet so once it's added it can probably be moved to its own profile.js file. I'd move the remaining stuff in index to the products.js file but there are some tests missing and it'd probably be better to put that in a separate pr.

oscarchong:
> > I think getUserProfile should go into "actions/users.js" as well just because it's user based and having all "users" in the same file would be more organized and "actions/index.js" can be all "products" stuff.
> 
> Alright. It would probably be more cluttered when there are actions added for editing the profile but we don't have that yet so once it's added it can probably be moved to its own profile.js file. I'd move the remaining stuff in index to the products.js file but there are some tests missing and it'd probably be better to put that in a separate pr.

okay that sounds good.

client/src/js/components/Login/Login.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from "react";
 import { Link } from "react-router-dom";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
-import { loginUser } from "../../actions/index";
+import { loginUser } from "../../actions/users";
```
synzen: hmm I notice you changed the imports from actions/users instead of actions/index - the way I did it for products was also move all the functions to the products actions file, and then in the 

https://github.com/CSC59939/QuikBorrow/blob/986def8ccd3f14c8a3137956fb8cf6e78abbd5c0/client/src/js/actions/index.js#L3

this allows us to directly import whatever was exported from products actions from index actions instead. which way do you think is better - import everything directly from index with this syntax, or explicitly import from the source file instead

Akasora39: Hmm that's a nice solution since it allows us to keep imports simple. Having it imported directly through an individual file would make it so they only have access to actions they will need but I'm not sure how much that matters. I think either way is fine.

Akasora39:
> Looks good, if we're gonna have a separate file for users actions, wouldn't it make sense to also have a separate file for the products actions?

Yeah we do. Like I said to Oscar, I was going to move the rest of the actions in index to the products actions and write tests for those since they were the only ones missing them but I'd rather make it a separate pr since I plan on making one for the test coverage we missed prior to agreeing to write tests for each file we made in each pr.

PR #147, opened 2019-12-11
added "add item" button and fix margin issue

oscarchong:
Before there's a UI issue on all the pages where there's extra white space when slide scrolled to the right. 

This PR fixed that issue. Additionally, I added an "Add Item" button on nav bar, this allows user to add/upload an item anywhere(when they're on product or profile page) besides just the homepage.

**Before:**
Screen Shot 2019-12-10 at 11 03 49 PM

**After:**
Screen Shot 2019-12-10 at 11 04 07 PM


Akasora39:
Do you know if this fixes it for all pages? Because I know the same issue happened for the detailed product page as well.

oscarchong:
> Do you know if this fixes it for all pages? Because I know the same issue happened for the detailed product page as well.

Yea, this will be fixed for all the pages

client/src/js/components/Navbar/SignedIn.test.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Enzyme, { shallow, mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { MemoryRouter } from 'react-router-dom';
+import { SignedIn } from './SignedIn';
```
Akasora39: Looks like you imported the unconnected component but you're connecting to a mock store so you probably want to either use the connected component or don't use a mock store.

oscarchong: oh alright i was trying out a lot of stuff but forgot to remove them

Akasora39:
> The test itself looks fine. I think it's a matter of how you imported the component.

Disregard this. Looks like it's moreso an issue regarding testing React hooks. I'll take a look at it.

PR #150, opened 2019-12-11
Set item price for AddItemModal

synzen:
This adds a price input to the AddItemModal


synzen:
> 
> 
> Geez that AddItemModal file is an absolute unit lol. Might want to see if we can move some stuff from there to other files like the axios call.

yeah, it's pretty dense :( the axios call can definitely be moved out of it as a redux action, for everything else though I'm not even sure. But no worries, it will be done in time for our international deployment!

PR #155, opened 2019-12-12
update Item rating

MicLi01:


server/controllers/items.js
```diff
@@ -53,10 +53,18 @@ const endRent = async (req, res, next) => {
       .catch(next);
 };
 
+/** @type {import('express').RequestHandler} */
+const updateRating = async (req, res, next) => {
+  itemServices.updateRating(req.params.itemId, req.body.rating)
```
Akasora39: Here `req.body.rating` should be `req.params.rating`.

server/services/items.js
```diff
@@ -66,6 +66,15 @@ const endRent = async (id) => {
   return updated;
 };
 
+/**
+ * @param {string} id - Item ID
+ * @returns {MongooseDocument} - The item with the modified rating
+ */
+const updateRating = async (id, rating) => {
+	const item = Item.findOneAndUpdate({'_id': id}, { '$set': {'rating': parseInt(rating) }}).exec();
```
Akasora39: As this is an async function, you should be using the await keyword which might have been why you weren't able to see the document being updated. In addition, findOneAndUpdate will return the old document unless provided with the option `{ new: true}`. I think you can get rid of the exec() call here.

server/services/items.unit.test.js
```diff
@@ -91,4 +91,13 @@ describe('services/item', function() {
       expect(returned).toEqual(updatedItem);
     });
   });
+  describe('updateRating', async function() {
```
Akasora39: You don't provide the async keyword to describe functions; just test and it functions require those.

server/services/items.unit.test.js
```diff
@@ -91,4 +91,13 @@ describe('services/item', function() {
...
+    it('finds the item by id', async function() {
+      const id = 'abc123';
+			const ratingData = 4;
+      Item.findOneAndUpdate.mockResolvedValueOnce({'exec': jest.fn()});
+      await itemServices.updateRating(id, ratingData);
+      expect(Item.findOneAndUpdate).toHaveBeenCalledWith({'_id': id}, { $set: {'rating': ratingData }});
```
Akasora39: I would change this so that you mock the resolved value of findOneAndUpdate to return an item with said id and ratingData. Then you can store what's returned from the itemServices.updateRating() call to a variable and expect what was returned to equal the item you mock returned.

server/services/items.unit.test.js
MicLi01: oops good catch

server/services/items.js
MicLi01: I didn't see that I missed the await keyword. With exec it still runs successfully though. I will change to await to make it consistent with tests.

server/controllers/items.js
MicLi01: Will change to params

server/services/items.unit.test.js
MicLi01: That makes sense will modify this test accordingly.

server/routes/api/items.js
```diff
@@ -51,4 +51,15 @@ router.get('/:itemId/endrent', [
...
 ], itemController.endRent);
 
+/**
+ * Updates rating of an item
+ *
+ * @memberof module:api/items
+ * @name GET /:itemId/updateRating/:rating/
+ */
+router.get('/:itemId/updateRating/:rating/', [
+  param('itemId').isAlphanumeric(),
```
Akasora39: Here you can probably use the custom isObjectId to ensure that it's a valid id.

server/services/items.unit.test.js
```diff
@@ -91,4 +91,16 @@ describe('services/item', function() {
...
+    it('finds the item by id', async function() {
+      const id = 'abc123';
+      const ratingData = 4;
+      Item.findOneAndUpdate.mockResolvedValueOnce({
+        'id': id,
+        'rating': ratingData,
+      });
+      const returned = await itemServices.updateRating(id, ratingData);
+      expect(returned).toEqual({'id': id, 'rating': ratingData});
+    });
```
yliu0715: You can destructure the `id` object by replacing `'id' : id` with just `id` on lines 99 and 103.

server/models/Item.js
```diff
@@ -6,6 +6,7 @@ const ItemSchema = new Schema({
   description: {type: String, default: ''},
   user: {type: Schema.Types.ObjectId, ref: 'User'},
   price: {type: Number, default: 5},
+  rating: {type: Number, default: 0},
```
yliu0715: Should probably give new accounts a default rating of 2.5 lol. 0 is kinda low.

Akasora39: This is for items and I think 0 fits because 0 can be considered as ambiguous since it means no ratings. Anything else would imply that there's been a rating already.

yliu0715: yes, you are right. My mistake, I thought it was Users lol

server/routes/api/items.js
MicLi01: Will make the change for the route for endrent as well

server/services/items.unit.test.js
MicLi01: Good to know. I replaced 'id': id with just id.

PR #157, opened 2019-12-15
Fix comments when not logged in and spacing

Akasora39:
Fixed bug where you could comment without being logged in. This would result in a comment being submitted but there was no username or id associated with it. Now users need to be logged in to comment.
Fixed spacing issue where the height of the grid for list of comments would equal the height of the grid above it causing long blocks of whitespace. Now each grid has its own height specified to prevent this.

image

### DIFF
--- a/client/src/js/components/Comments/Comment.js
+++ b/client/src/js/components/Comments/Comment.js
@@ -14,7 +14,8 @@ export class Comment extends React.Component {
 		super();
 		this.state = {
 			body: "",
-			rating: 5
+			rating: 5,
+			error: ""
 		};
 	}
 	/** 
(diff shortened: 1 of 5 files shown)


yliu0715:
@Akasora39 Can you provide an image to show the changes on the spacing?

Akasora39:
> @Akasora39 Can you provide an image to show the changes on the spacing?

Added  to description

client/src/js/components/Comments/Comment.js
```diff
@@ -33,20 +34,28 @@ export class Comment extends React.Component {
...
+			this.setState({ error: "You must be logged in to comment."})
 		}
-		this.props.createComment(data);	
-		window.location.reload(false);	
 	}
 	render()
 	{
 		return(
 			<form noValidate onSubmit={this.onSubmit}>
+				<span>{this.state.error}</span>
```
yliu0715: Looks good, guess you can also add in red text for better UI like:
```suggestion
				<span style="color:red">{this.state.error}</span>
```

or maybe wrap the span using styled components like we've been doing

PR #158, opened 2019-12-16
Transactions On FrontEnd

yliu0715:
- transactions list has been added on frontend
- a couple unit tests for transactions are added

transactions

Each transaction has a link to the corresponding item associated with that transaction.

Note: The transaction didn't have an item id field, probably cuz it was used for testing.


client/src/js/actions/products.test.js
```diff
@@ -134,4 +134,41 @@ describe('actions/products', () => {
...
+        beforeEach(() => {
+            mock = jest.spyOn(axios, 'post');
+        });
+        afterEach(() => {
+            mock.mockRestore();
+        });
+
+        it('should fetch a user\'s transactions', async () => {
+            const dispatch = jest.fn();
+            mock.mockResolvedValue();
```
Akasora39: You aren't mocking a resolved value so when your action expects the payload to be res.data, it's getting undefined instead which is causing the error.

yliu0715: Oh, nice catch lol

PR #159, opened 2019-12-16
Rating of product now reflects ratings in comments

Akasora39:
Ratings now reflect the ratings in comments. 
Added updateRating action to call api route.
Updated Comment to use getComments and updateRating actions. I feel like these additions are a bit excessive for now but I think this makes the most sense since these events should only occur once a comment has been successfully submitted.

Also updated default images. (Don't tell Michael shhhhhhhh)


synzen:
a student website without any memes?!



kidding lol

client/src/js/components/Comments/Comment.js
```diff
@@ -18,6 +19,15 @@ export class Comment extends React.Component {
...
 	}
+
+	getRatings(comments)
+	{
+		let total = 0;
+		comments.map(comment => {
+			total+=comment.rating;
+		});
+		return total;
+	}
```
yliu0715: I think it's more appropriate to use the `reduce()` method here to accumulate the sum like this in 1 line: `return comments.reduce((total, comment) => total + comment.rating, 0)`. The first argument defines what operation you want to compute on the array and accumulates the total and the second argument is the value of the current iteration. Remember them troeger stuff? I think its similar to the accumulate function in scheme

Akasora39: Yeah you're right. Couldn't remember what the function was called at the time but this was what I was trying to accomplish. No lint warnings now 👍

PR #160, opened 2019-12-18
Fix tiny error causing tests to fail

Akasora39:
The smallest change that was necessary because I didn't want to see an incredibly long import statement lol.
This is what we sign up for when writing tests regarding a core aspect of our application.


Akasora39:
> funny, I didn't even notice master had widespread failures lol

Same lol. Saw tests passing on my branch and merged and the checks on master take a while so I just closed out thinking they passed.

